PR #24, opened 2017-02-20
feat: Create keyboard listeners

mingard:
feat: Basic keyboard event lib

@eduardoboucas Still needs:
- [ ] OS variants
- [ ] Shortcut shortcuts e.g. `keyboard.DELETE`
- [x] `off` method
- [x] map all keys

### DIFF
--- a/dadi/frontend/lib/keyboard.js
+++ b/dadi/frontend/lib/keyboard.js
@@ -1,0 +1,107 @@
+'use strict'
+
+class Keys {
+
+  get keys () {
+    return [
+      {
+        key: ' ',
+        code: 'Space'
+      },
+      {
+        key: 'a',
+        code: 'KeyA'
+      }
+    ]
+  }
+
+  find (pattern) {
+    return pattern.map(key => {
+      return this.keys.find(local => {
+        return local.key.toLowerCase() === key || local.code.toLowerCase() === key
+      })
+    }).filter(key => {
+      return key
+    })
+  }
+}
+
+class Pattern {
+  constructor (pattern, keys) {
+    this.pattern = pattern
+    this.keys = keys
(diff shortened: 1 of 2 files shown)


dadi/frontend/lib/keyboard.js
```diff
@@ -0,0 +1,176 @@
...
+  constructor (pattern, keys) {
+    this.pattern = pattern
+    this.keys = keys
+    this.active = 0
+  }
+
+  next (keyCode) {
+    if (this.match(keyCode, this.keys[this.active])) {
+      if (this.active < this.keys.length -1) {
+        this.active ++
```
eduardoboucas: I'd prefer `this.active++` (without the space)

dadi/frontend/lib/keyboard.js
```diff
@@ -0,0 +1,176 @@
...
+    return this.shortcuts.find(shortcut => {
+      return shortcut.next(keyCode)
+    })
+  }
+
+  on (pattern) {
+    let keys = this.find(pattern.toLowerCase().split('+'))
+    let shortcut = new Pattern(pattern, keys)
+    this.shortcuts.push(shortcut)
+    return shortcut
```
eduardoboucas: It'd be great to have more consistency around spacing (perhaps something we could add to the doc). I think having a blank line after the declarations bit of every block (so after the two `let`) and before the `return` would make things more readable. You've done that in some parts of the PR.

mingard: @eduardoboucas should be better now

dadi/frontend/lib/keyboard.js
mingard: Done

PR #40, opened 2017-02-28
Submit SignIn on Return and break out Style

eduardoboucas:


dadi/frontend/components/DropdownItem/DropdownItem.jsx
```diff
@@ -2,17 +2,11 @@
...
-import { Style } from 'lib/util'
+import Style from 'lib/Style'
 import styles from './DropdownItem.css'
 
 export default class DropdownItem extends Component {
   handleClick(event) {
-
-    // Don't stop propagation for anchor node
-    if (!this.props.href) {
-      event.stopPropagation()
```
mingard: @eduardoboucas what have you done instead to stop this breaking Router?

eduardoboucas: Erm, I didn't mean to remove this. Sorry, I'll fix tomorrow.

mingard: re-ping - this part of the next batch of updates from you?

eduardoboucas: Yes, but you can still merge this PR. It was the presence of `stopPropagation()` (which I removed) that was breaking the router. In fact, I'm not sure we need it anymore, so all should be good.

mingard: Ah, excellent. Then no need to add this back then? Merged

eduardoboucas: Not for now. I'll dig deeper once I revisit the `ButtonWithOptions` component.

Issue #41, opened 2017-02-28
Switch to `defaultProps` in all components


mingard:
New ES6 syntax for default props

```javascript
static defaultProps = {
  type: 'button'
}
```

eduardoboucas:
Closed by #44.

PR #44, opened 2017-03-02
Too many things

eduardoboucas:
This PR:

- Adds user elements to `Header`
- Adds WIP sign out process (missing the session destroy logic on the app side)
- Changes User store to store just `null` when no user is signed in
- Adds error state to `Label`
- Add `Table` and `SyncTable` components (and sub-components)
- Add new document list view


eduardoboucas:
@mingard I've removed the link from the *Sign out* button, as per our conversation.

I've also removed prop fallbacks in favour of `defaultProps` in the rest of the components, so this should close #41.

dadi/frontend/lib/session.js
```diff
@@ -8,7 +8,26 @@ import 'unfetch'
 const Session = function () {}
 
 /**
+ * Builds a user object from the app endpoint response.
+ *
+ * @param {object} The response from the app endpoint.
+ *
+ * @return {object} The user object.
+ */
+Session.prototype._buildUserObject = function (user) {
```
mingard: @eduardoboucas underscore as an local scope identifier? I don't agree with using an unsupported convention.

eduardoboucas: Changed.

PR #51, opened 2017-03-03
Add selection to document list view

eduardoboucas:


dadi/frontend/components/Table/Table.jsx
```diff
@@ -16,16 +16,89 @@ export default class Table extends Component {
     selectable: true
   }
 
+  constructor(props) {
+    super(props)
+
+    this.state.selectedRows = {}
```
mingard: @eduardoboucas Aren't we using `static defaultProps` now? Why are we setting `state` outside of an action?

eduardoboucas: We are using `defaultProps` when we have props, this is state. As for using state outside of an action, we discussed that before and we've been doing that in some components that should be able to manage their own state independently of the store (e.g. error state in a label).

In this case, we're talking about a table being able to know which of its rows are selected. It doesn't know anything about documents or collections, but it does need to have the concept of selected rows. Otherwise you wouldn't be able to have a table that doesn't directly represent an entity in the store.

In this particular case, it might make sense to also connect it to the store, so that other components in the application know which documents are selected in the document list view — but that should be specific to this particular case, not something that we enforce for all instances. And even then, the table should still keep its own state, but it also communicate it to (and take initial state from) the rest of the application.

PR #61, opened 2017-03-07
feat: generate filters

mingard:


dadi/frontend/components/DocumentFilter/DocumentFilter.jsx
```diff
@@ -20,29 +25,63 @@ export default class DocumentFilter extends Component {
...
+          <option disabled selected value>Select a type</option>
           {Object.keys(this.filterTypes).map(key => (
-            <option name={key}>{this.filterTypes[key]}</option>
+            <option selected={this.state.type === key} key={key} value={key}>{this.filterTypes[key]}</option>
           ))}
         </select>
-        {field && value && (
-          <TextInput value={this.filterValue(value)} />
+        {this.state.field && this.state.type && (
+          <TextInput id="value" value={this.state.value} onChange={this.onChange.bind(this)} />
```
eduardoboucas: You're passing a static string as the `id` prop of a `TextInput`. This means that. if you have multiple of these, you'll have multiple DOM elements with the same id. What's the id for here?

eduardoboucas: Is it so that you can use `onChange` with different elements? If so, I'd suggest using the second argument of `.bind()`, as used [here](https://github.com/dadi/publish/blob/d0c5b545785a9115ed3634c55e34befce8d01930/dadi/frontend/components/Header/Header.jsx#L78), which will then come through as the first argument of `onChange`.

dadi/frontend/components/DocumentFilters/DocumentFilters.jsx
```diff
@@ -5,29 +5,68 @@ import {h, Component} from 'preact'
...
+          <Button onClick={this.addFilter.bind(this)}>Add</Button>
+        )}
       </div>
     )
   }
 
+  updateUrlFilters(filter) {
+    const {updateUrlParams} = this.props
+    const {filters} = this.state
+    this.setState(Object.assign(filters, filter))
```
eduardoboucas: minor: space after initialisation (after 54)

dadi/frontend/views/DocumentList/DocumentList.jsx
```diff
@@ -41,7 +41,7 @@ class DocumentList extends Component {
 
     return (
       <section class="Documents">
-        <DocumentFilters filter={filter} collection={collection} />
+        <DocumentFilters filter={filter} collection={collection} updateUrlParams={this.updateUrlParams.bind(this)} />
```
eduardoboucas: minor: I'd personally break this into multiple lines.

dadi/frontend/components/DocumentFilter/DocumentFilter.jsx
mingard: @eduardoboucas id property was because that's all I could see available in the components standard props

dadi/frontend/views/DocumentList/DocumentList.jsx
mingard: @eduardoboucas will do

dadi/frontend/components/DocumentFilter/DocumentFilter.jsx
```diff
@@ -20,29 +25,66 @@ export default class DocumentFilter extends Component {
...
       </div>
     )
   }
 
+  onChange(elementId, event) {
+    const {updateFilter} = this.props
+      
+    this.setState({[elementId]: event.target.value})
+    // {!} TO-DO add further evaluation in `Object.js`
+    if (!Object.keys(this.state).filter(key => {
```
eduardoboucas: I find this `if` statement a bit hard to read. Could you use an intermediate variable with a descriptive name to help understand what's going on?

dadi/frontend/components/DocumentFilters/DocumentFilters.jsx
```diff
@@ -5,29 +5,70 @@ import {h, Component} from 'preact'
...
     return (
       <div>
-        {filters && Object.keys(filters).length && collection && (
+        <Button onClick={this.toggleFilters.bind(this)}>Filters</Button>
+
+        {visible && filters && Object.keys(filters).length && collection && (
           <div>
             {Object.keys(filters).map(key => ( 
-              <DocumentFilter field={key} value={filters[key]} fields={collection.fields} />
+              <DocumentFilter field={key} value={filters[key]} fields={collection.fields} updateFilter={this.updateUrlFilters.bind(this)} />
```
eduardoboucas: I'd also break this into multiple lines. 😇

dadi/frontend/components/DocumentFilters/DocumentFilters.jsx
```diff
@@ -5,29 +5,70 @@ import {h, Component} from 'preact'
...
+        )}
       </div>
     )
   }
 
+  updateUrlFilters(filter) {
+    const {updateUrlParams} = this.props
+    const {filters} = this.state
+
+    this.setState({filters: Object.assign(filters || {}, filter)})
```
eduardoboucas: If `filters` exists and you do `Object.assign(filters, something)`, you're assigning something directly to a state variable, which we shouldn't do.

Why not just:

```js
this.setState({filters: Object.assign({}, filters, filter)})
```

It'll be fine even if `filters` is `null`.

PR #66, opened 2017-03-09
feat: routing method with query params

mingard:
This PR addresses the lack of support for query parameter changing with `preact-router`. 

Using the `router` method from `lib/router` you can now call basic routes as before, as well as update or override url parameters.

### Update 

```javascript

// Basic route change
router({path: '/'})

// Change parameters
// Note that path will default to `window.location.pathname`
router({params: {foo: "bar"}})

// Append existing parameters with `Object.assign`
router({params: {foo: "bar", update: true}})

// Clear all parameters in the current route (not that useful)
router()
```

The other minor change is that now `state.router` has an extra parameter, `params`. This stores an Object of the query parameters:

```javascript
{
  foo: "bar",
  filters: {
    email: "publish_wizard@dadi.co"
  }
}
```

### DIFF
--- a/dadi/frontend/lib/router.js
+++ b/dadi/frontend/lib/router.js
@@ -1,0 +1,22 @@
+'use strict'
+
+import {route} from 'preact-router'
+import {urlHelper} from 'lib/util'
+
+export function router({path=window.location.pathname, params=null, update=false}) {
+  let newParams
+
+  if (params) {
+    if (update) {
+      let currentParams = urlHelper().paramsToObject(window.location.search)
+      newParams = urlHelper().paramsToString(Object.assign({}, currentParams, params))
+    } else {
+      newParams = urlHelper().paramsToString(params)
+    }
+  }
+
+  route(`${path}${decodeURIComponent(newParams) || ''}`)
+  return {
+
+  }
+}
--- a/dadi/frontend/lib/router.js
+++ b/dadi/frontend/lib/router.js
@@ -16,7 +16,4 @@ export function router({path=window.location.pathname, params=null, update=false
   }
 
   route(`${path}${decodeURIComponent(newParams) || ''}`)
-  return {
-
-  }
 }
--- a/dadi/frontend/middleware/router.js
+++ b/dadi/frontend/middleware/router.js
@@ -26,7 +26,6 @@ export default function syncRouteWithStore(history, store, {
   }
 
   const handleLocationChange = (location) => {
-    console.log("handleLocationChange")
     if (isTimeTraveling) {
       return
     }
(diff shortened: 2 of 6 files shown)


dadi/frontend/lib/router.js
```diff
@@ -0,0 +1,22 @@
...
+    if (update) {
+      let currentParams = urlHelper().paramsToObject(window.location.search)
+      newParams = urlHelper().paramsToString(Object.assign({}, currentParams, params))
+    } else {
+      newParams = urlHelper().paramsToString(params)
+    }
+  }
+
+  route(`${path}${decodeURIComponent(newParams) || ''}`)
+  return {
```
eduardoboucas: Is something missing here?

dadi/frontend/middleware/router.js
```diff
@@ -25,6 +26,7 @@ export default function syncRouteWithStore(history, store, {
   }
 
   const handleLocationChange = (location) => {
+    console.log("handleLocationChange")
```
eduardoboucas: Remove this?

dadi/frontend/middleware/router.js
```diff
@@ -51,7 +56,7 @@ export default function syncRouteWithStore(history, store, {
       return
     }
     isTimeTraveling = true
-    history.push(locationInStore)
+    // history.push(locationInStore)
     isTimeTraveling = false
```
eduardoboucas: Setting to `true` and then setting to `false` straight away?

dadi/frontend/views/DocumentList/DocumentList.jsx
```diff
@@ -164,13 +165,14 @@ class DocumentList extends Component {
...
     }).catch((err) => {
       actions.clearDocumentList()
-
       // TODO: Graceful deal with failure
     })
   }
 
   updateUrlParams (filters) {
-    // {!} TO-DO update url filters
+    const {actions, state} = this.props
```
eduardoboucas: Could you add a blank line after initialisation, please?

dadi/frontend/lib/router.js
mingard: @eduardoboucas yes, I'm adding some methods in the return

dadi/frontend/middleware/router.js
mingard: ah yes, will do!

dadi/frontend/middleware/router.js
mingard: @eduardoboucas yes that's required, although that should not be commented out

dadi/frontend/views/DocumentList/DocumentList.jsx
mingard: done

PR #68, opened 2017-03-10
feat: build filters

mingard:


dadi/frontend/components/DocumentFilters/DocumentFilters.jsx
```diff
@@ -19,52 +19,75 @@ export default class DocumentFilters extends Component {
     } = this.props
 
     // Evaluate passed filter / store in state
+    let filters = (isValidJSON(filter) ? objectToArray(JSON.parse(filter), 'field') : []).map(this.deconstructFilters)
     this.state = {
```
eduardoboucas: We shouldn't be mutating `state` directly, only with `setState()`.

dadi/frontend/components/DocumentFilters/DocumentFilters.jsx
```diff
@@ -19,52 +19,75 @@ export default class DocumentFilters extends Component {
...
         )}
       </div>
     )
   }
 
-  updateUrlFilters(filter) {
+  deconstructFilters(filter) {
+    // If there is no filter type and the field value is a string, add a type
+    if (typeof filter.value === 'string') return Object.assign(filter, {type: '$eq'})
+    else {
```
eduardoboucas: This looks a bit odd. We usually put the body of the `if` on a new line if there is an `else` block, like:

```js
  deconstructFilters(filter) {
    // If there is no filter type and the field value is a string, add a type
    if (typeof filter.value === 'string') {
      return Object.assign(filter, {type: '$eq'})
    } else {
      let type = Object.keys(filter.value)[0]

      return Object.assign(filter, {type: type, value: filter.value[type]})
    }
  }
```

dadi/frontend/components/DocumentFilters/DocumentFilters.jsx
mingard: @eduardoboucas sure, but this isn't

mingard: This is in pre-render. Just like https://github.com/dadi/publish/blob/07adcb74e9766f3fb1bc85894dec675bf4565121/dadi/frontend/components/FieldImage/FieldImage.jsx#L13-L16

mingard: I have updated this, but suggest we do it everywhere it we really want to treat state as immutable

PR #74, opened 2017-03-13
feat: error handling with messages

mingard:


dadi/frontend/views/SignIn/SignIn.jsx
```diff
@@ -82,6 +82,10 @@ class SignIn extends Component {
     )
   }
 
+  errors() {
```
eduardoboucas: What's this for?

mingard: ah, meant to remove this

PR #112, opened 2017-03-21
feat: handle image upload

mingard:
WIP!

feat: image data updated with onChange
feat: image data saving (currently as raw)
feat: reference fields located in document, ready to use with reference document create/update method

### DIFF
--- a/frontend/components/FieldImage/FieldImage.jsx
+++ b/frontend/components/FieldImage/FieldImage.jsx
@@ -1,6 +1,7 @@
 'use strict'
 
 import {h, Component} from 'preact'
+import proptypes from 'proptypes'
 import 'fetch'
 
 import {buildUrl} from 'lib/router'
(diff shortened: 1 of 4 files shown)


frontend/components/FieldImage/FieldImage.jsx
```diff
@@ -9,27 +10,54 @@ import Label from 'components/Label/Label'
 
 export default class FieldImage extends Component { 
 
+  static propTypes = {
+
+    maxWidth: proptypes.number,
```
eduardoboucas: Do we need `maxWidth` as a prop? We talked about all field components receiving the exact same props, so it's easy to make their loading dynamic.

If it does need it, can you add documentation to the prop?

mingard: @eduardoboucas it's temporary. I'll make it static for now so that later it can be handled with css

PR #115, opened 2017-03-22
Add ESLint

eduardoboucas:
This PR adds [ESLint](http://eslint.org/). I've included the rules that I think we should use. They are all listed in `.eslintrc.json`.

To see what each of the rules do, go to `http://eslint.org/docs/rules/NAME-OF-THE-RULE` or see [the list of available rules](http://eslint.org/docs/rules/).

ESLint has a `--fix` flag that can, for most of the rules, change any files that don't adhere to them. Once we're happy with the list of rules, I can do that and also fix any files that need to be changed manually.

I also suggest we add a [Travis](https://travis-ci.org/) job, which will lint every pull request automatically and flag any discrepancies. I can take care of that once we're happy with this PR.


.eslintrc.json
```diff
@@ -0,0 +1,142 @@
...
+                "allowArrayStart": true
+            }
+        ],
+        "lines-around-directive": [
+            "error",
+            "always"
+        ],
+        "max-len": [
+            "error",
+            80
```
eduardoboucas: 80 chars is the default and a somewhat consensual value used as the maximum line length. May be too strict for us? Should we do 90 or 100?

mingard: 80 seems okay

PR #121, opened 2017-03-23
Feat/lazy load assets

mingard:
In this PR:
feat: make `FieldImage` a more generic `FieldAsset`
feat: start styling `FieldAsset`
feat: introduce a `LazyLoader` container (WIP)

### DIFF
--- a/frontend/components/FieldAsset/FieldAsset.css
+++ b/frontend/components/FieldAsset/FieldAsset.css
@@ -1,0 +1,28 @@
+.container {
+  width: 100%;
+  margin-top: 9px;
+  height: 40px;
+  line-height: 25px;
+  font-size: 15px;
+  box-sizing: border-box;
+  border: 1px solid #e4e4e4;
+  border-radius: 5px;
+  outline: 0;
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+}
+
+.container[preview] {
+  background-color: #fafafa;
+  height: auto;
+  max-height: 130px;
+}
+
+.image-thumb {
+  height: auto;
+  max-height: 110px;
+  width: auto;
+  max-width: 50%;
+  margin: 10px;
+}
--- a/frontend/components/FieldAsset/FieldAsset.css
+++ b/frontend/components/FieldAsset/FieldAsset.css
@@ -13,7 +13,7 @@
   align-items: center;
 }
 
-.container[preview] {
+.show-preview {
   background-color: #fafafa;
   height: auto;
   max-height: 130px;
--- a/frontend/containers/LazyLoader/LazyLoader.jsx
+++ b/frontend/containers/LazyLoader/LazyLoader.jsx
@@ -1,0 +1,61 @@
+'use strict'
+
+import {h, Component} from 'preact'
+import {connect} from 'preact-redux'
+import proptypes from 'proptypes'
+import {bindActionCreators} from 'redux'
+import {connectHelper} from 'lib/util'
+
+import * as appActions from 'actions/appActions'
+
+class LazyLoader extends Component {
+  static propTypes = {
+
+    /**
+     * Load only when app status is idle
+     */
+    idleOnly: proptypes.bool,
+
+    /**
+     * The placeholder node to be rendered before load
+     */
+    placeholder: proptypes.node,
+    /**
+     * The asset to be loaded when appropriate
+     */
+    children: proptypes.node
+  }
+
+  static defaultProps = {
+    idleOnly: true
+  }
+
+  render() {
+    const {children} = this.props
+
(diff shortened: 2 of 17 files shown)


frontend/components/FieldAsset/FieldAsset.css
```diff
@@ -0,0 +1,28 @@
...
+  box-sizing: border-box;
+  border: 1px solid #e4e4e4;
+  border-radius: 5px;
+  outline: 0;
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+}
+
+.container[preview] {
```
eduardoboucas: We should only style by class name, not attribute. If you want to add a styling block based on whether a condition is met, you should use the `Style` class. See [this](https://github.com/dadi/publish/blob/master/frontend/components/Button/Button.jsx#L99) as an example.

frontend/containers/LazyLoader/LazyLoader.jsx
```diff
@@ -0,0 +1,60 @@
...
+import {h, Component} from 'preact'
+import {connect} from 'preact-redux'
+import proptypes from 'proptypes'
+import {bindActionCreators} from 'redux'
+import {connectHelper} from 'lib/util'
+
+import * as appActions from 'actions/appActions'
+
+class LazyLoader extends Component {
+  static propTypes = {
```
eduardoboucas: Could you follow the same spacing convention as used in the proptypes definition of other components? (Also period at the end of the sentences, if possible)

frontend/containers/LazyLoader/LazyLoader.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+    const canLoad = this.evalLoadingConditions()
+
+    if (!canLoad) {
+      if (placeholder) return (placeholder)
+
+      return null
+    }
+
+    return (<div>{children}</div>)
```
eduardoboucas: We've been doing:

```jsx
return (
  <div>{children}</div>
)
```

... even for components that return a single element.

frontend/components/FieldAsset/FieldAsset.css
mingard: @eduardoboucas I was just following your lead

eduardoboucas: I don't follow.

mingard: You wrote that

frontend/containers/LazyLoader/LazyLoader.jsx
mingard: Sure

frontend/components/FieldAsset/FieldAsset.css
eduardoboucas: `multiple` is a standard HTML attribute that identifies `<select>` elements with multiple options. `preview` is something we made up.

mingard: > We should only style by class name, not attribute
Unless we're being consistent on this, I don't see merit in this request

frontend/containers/LazyLoader/LazyLoader.jsx
mingard: This isn't showing as resolved but as far as I can see it is!

frontend/components/FieldAsset/FieldAsset.css
eduardoboucas: You're adding a non-standard attribute to the element, in the actual rendered DOM, without it even being prefixed with `data-`. There's no reason to do this when all you want to do is target it for styling.

That isn't the case in `FieldString`, but I can change it to also use a class if that makes you happier.

mingard: Yes, I think we need to be a bit more consistent here. Do that. Add to style guide too if you've the time.

eduardoboucas: It is in the style guide: https://github.com/dadi/publish/blob/master/docs/styling.md#21-target-class-names

eduardoboucas: As well as documentation for the `Style` class, the solution I proposed: https://github.com/dadi/publish/blob/master/docs/styling.md#31-the-style-class

mingard: Great. lets enforce it

mingard: I think everything else there is covered

eduardoboucas: I've updated `FieldString`. Can you update yours?

mingard: Done

PR #131, opened 2017-03-27
Add DateTime and DocumentEditToolbar components

eduardoboucas:


frontend/views/Home/Home.jsx
```diff
@@ -3,27 +3,16 @@ import {connect} from 'preact-redux'
...
-    const {state} = this.props
+    const d = new Date(2015, 2, 10, 5, 30, 20)
 
     return (
-      <div>
-        <Label label="Title" comment="Optional">
-          <TextInput />
-        </Label>
-      </div>
+      <DateTime date={d} />
```
eduardoboucas: This was just so that I could easily test the `DateTime` component, since we're not really doing anything with this view yet. What it should it be, btw?

frontend/components/DateTime/DateTime.jsx
```diff
@@ -0,0 +1,220 @@
...
+const WEEK = 7 * DAY
+const YEAR = 365 * DAY
+const MONTH = YEAR / 12
+
+/**
+ * The time elapsed since a given event, as an automatically updated string.
+ */
+export default class ElapsedTime extends Component {
+  static propTypes = {
+    /**
```
mingard: Spacing

frontend/components/DocumentEditToolbar/DocumentEditToolbar.jsx
```diff
@@ -0,0 +1,127 @@
...
+     */
+    method: proptypes.oneOf(['edit', 'new']),
+
+    /**
+     * A callback to be fired when the "Save" button is pressed.
+     */
+    onSave: proptypes.func,
+
+    /**
+     * The document currently being edited.
```
mingard: Incorrect descriptions

frontend/containers/DocumentEdit/DocumentEdit.jsx
```diff
@@ -235,28 +222,13 @@ class DocumentEdit extends Component {
...
-            >Delete</Button>
-            <PeerList
-              peers={document.peers}
-            />
-          </div>
-
-          <div>
-            <ButtonWithOptions
-              accent="save"
-              disabled={hasValidationErrors || hasConnectionIssues}
```
mingard: is `hasConnectionIssues` still being used? Doesn't appear to be

frontend/components/DocumentEditToolbar/DocumentEditToolbar.jsx
```diff
@@ -0,0 +1,127 @@
...
+                relative={true}
+              />
+            </p>
+          </div>
+        </div>
+
+        <div class={styles['padded-group']}>
+          <ButtonWithOptions
+            accent="save"
+            disabled={hasValidationErrors}
```
mingard: missing the `hasConnectionIssues` variable

frontend/containers/DocumentEdit/DocumentEdit.jsx
eduardoboucas: I removed it accidentally when doing the merge. It has been reinstated now.

frontend/components/DateTime/DateTime.jsx
eduardoboucas: What about it?

frontend/components/DocumentEditToolbar/DocumentEditToolbar.jsx
eduardoboucas: Done

frontend/containers/DocumentEdit/DocumentEdit.jsx
mingard: Okay

frontend/components/DateTime/DateTime.jsx
mingard: Spacing is required about the comments

frontend/views/Home/Home.jsx
mingard: Good question! Lets raise a ticket and discuss there

frontend/components/DocumentEditToolbar/DocumentEditToolbar.jsx
```diff
@@ -39,19 +39,23 @@ export default class DocumentListToolbar extends Component {
...
+    disabled: false
+  }
+
   constructor(props) {
     super(props)
   }
 
   render() {
     const {
+      disabled,
```
mingard: @eduardoboucas why are we consolidating both validationErrors and connectionErrors here? Why would we stop deleting if the document had validation errors?

frontend/components/DateTime/DateTime.jsx
eduardoboucas: If you look at the other component (except perhaps a few that have been missed recently), we don't use that spacing. That is consistent with our linting rules that determine that comments don't need extra lines when at the beginning of a block or object literal (see [here](https://github.com/dadi/publish/blob/master/.eslintrc.json#L68-L70)).

frontend/components/DocumentEditToolbar/DocumentEditToolbar.jsx
eduardoboucas: That is very true, I didn't think of that. Will change.

frontend/components/DateTime/DateTime.jsx
mingard: So the first property has no spacing before comments but all subsequent prop comments do? Seems strange IMO, but fine.

frontend/components/DocumentEditToolbar/DocumentEditToolbar.jsx
eduardoboucas: Does it make sense now?

frontend/components/DateTime/DateTime.jsx
eduardoboucas: The idea is to have spacing between each comment/property pair. The first one doesn't need it because it's at the top of the block.

mingard: Okay

frontend/components/DocumentEditToolbar/DocumentEditToolbar.jsx
```diff
@@ -111,7 +117,7 @@ export default class DocumentListToolbar extends Component {
         <div class={styles['padded-group']}>
           <ButtonWithOptions
             accent="save"
-            disabled={disabled}
+            disabled={hasValidationErrors || disabled}
```
mingard: I'd prefer we're as explicit about connection errors as we are about validation. `disabled` should be `hasConnectionErrors`

eduardoboucas: My idea to use `disabled` was to keep it generic so that parents can choose to disable the controls in the toolbar for other reasons if they wish to. We're using `hasConnectionIssues` in `DocumentEdit` and being explicit about that, I don't think the toolbar necessarily needs to be aware of the concept of connection issues. It just enables or disables the controls based on what the parent tells it.

Your call, though.

mingard: That logic could be applied to `hasValidationErrors` too, as both simply disable the component. I think it makes sense to be clear on why the condition is true so would rather we keep both well labelled.

mingard: Also, deciding whether `hasConnectionIssues` is used needs to be clear. It should be reusable and therefore explicit

Issue #128, opened 2017-03-27
Store document state in localstorage


mingard:
In order to preserve the state of 'draft' documents in post-mvp, we need to replicate the local state value into localstorage. For now, we just need to store the value.

I suggest we use https://github.com/elgerlambert/redux-localstorage but happy if we roll our own.

eduardoboucas:
A couple of questions about this:

1. Do we want to show a warning saying "You have unsaved changes" if the state of the document when the user loads the page comes from a locally-saved version and not the remote API? If so, how would this look visually? Sort of relates to #86.

1. What happens if the remote document has changed since you last opened it, meaning that your local version has potential conflicts with the remote? Do we want to try some form of reconciliation (might be tricky?), or do we simply warn the user and let them make a decision?

mingard:
1. Yes, and that probably means #86 should have a more versatile error message as you suggest
2. That's a good question. We probably don't want to start down this path yet, as it's likely to be fairly complex. Is there anything about this decision that would block you now?

eduardoboucas:
> Yes, and that probably means #86 should have a more versatile error message as you suggest

Indeed. Will discuss on that ticket.

> Is there anything about this decision that would block you now?

Not particularly. I can start with a simple approach of flagging to the user when the remote is out-of-sync and then we can think of a more comprehensive solution (possibly even post-MVP?).

mingard:
@eduardoboucas 
> I can start with a simple approach of flagging to the user when the remote is out-of-sync
Sounds good to me

eduardoboucas:
Closed via #139.

PR #139, opened 2017-03-29
Add notification centre and local storage

### LINKED ISSUE #86: Toaster UI
From Slack:

>I’m looking at the toaster/info banner (http://cloud.dadi.technology/kpk3). couple of questions:
>
>1. in your mock up, it pushes down the content of the page. is this intended?
>2. it’s positioned with the header, which is not fixed positioned. this means that if I have scrolled down the page and hit save, I won’t be able to see the toaster unless I scroll all the way up again. what are your thoughts?
>3. does it automatically close itself after a period of time, or do you have to close it? the label saying “10 minutes ago” makes me thing the latter, but I don’t see any controls.

eduardoboucas:
Closes #86 and #128.


frontend/containers/NotificationCentre/NotificationCentre.jsx
```diff
@@ -0,0 +1,147 @@
...
+      callback()
+    }
+
+    this.dismiss()
+  }
+}
+
+export default connectHelper(
+  state => state,
+  dispatch => bindActionCreators({...appActions}, dispatch)
```
mingard: Change to `dispatch => bindActionCreators(appActions, dispatch)`

frontend/reducers/document.js
```diff
@@ -1,40 +1,96 @@
...
 
+// Boolean fields are a bit special. Any required field that hasn't
+// been touched by the user in the UI should generate a validation error,
+// except for Boolean fields, as these don't have a "neutral" state.
+// You can start editing a document and see a Boolean field set to
+// "false" and that may be exactly how you want it, but at that point the
+// field doesn't exist in the document store yet because there wasn't a
+// change event. To get around this, when we create our local copy of
+// the document we check for any Boolean fields in the collection
+// that aren't in the document object and set those to `false`.
```
mingard: Is it worth considering a `neutral` state? https://css-tricks.com/indeterminate-checkboxes/ style

frontend/containers/NotificationCentre/NotificationCentre.jsx
eduardoboucas: Done.

frontend/reducers/document.js
eduardoboucas: I don't think so. That is purely visual, in case we want to display a neutral state to the user, which I believe we don't.

> The indeterminate state is visual only. The checkbox is still either checked or unchecked as a state. That means the visual indeterminate state masks the real value of the checkbox, so that better make sense in your UI!

mingard: I don't think we need to display it, but would make sense to apply it to the component so that on save, the value is not set

eduardoboucas: I don't understand how this helps us with the problem addressed in the comment. If you have a Boolean field that's required, you can't save a document without including it in the document body. For that to happen, we need to make sure we append that field to the document even if the user hasn't interacted with the UI element, which is precisely what this piece of logic is doing for Boolean fields.

mingard: Okay, so if the field _isn't_ required and the user doesn't interact with the component, setting the value to true/false, it won't post anything?

eduardoboucas: At the moment it's not checking to see if the field is required, but it's something I can add pretty easily.

mingard: That's what I was suggesting.

eduardoboucas: Done!

mingard: Ninja

PR #159, opened 2017-04-04
fix: socket updates on enter, skipping current user

mingard:


frontend/components/DocumentEditToolbar/DocumentEditToolbar.jsx
```diff
@@ -76,7 +76,7 @@ export default class DocumentListToolbar extends Component {
 
     return (
       <Toolbar>
-        {peers && peers.length &&
+        {Boolean(peers && peers.length) &&
```
eduardoboucas: Why the `Boolean()`?

mingard: When `peers` is length 0 the condition `peers.length` is still a valid condition, meaning the output in the dom is just `0`.

eduardoboucas: Ah okay. I think `peers && (peers.length > 0)` would be more descriptive, but I guess that's just personal taste.

mingard: I'll change to that

PR #160, opened 2017-04-05
style: remove extra lb

mingard:
Ready to be used with bulk actions


frontend/containers/DocumentList/DocumentList.jsx
```diff
@@ -75,7 +75,7 @@ class DocumentList extends Component {
   constructor(props) {
     super(props)
 
-    this.state.selectedRows = {}
+    this.state = this.defaultLocalState()
```
eduardoboucas: I still don't see the need for this to be a function. I'd personally define the default state as an object within the constructor, which is where people expect it, and use that as the argument for `setState()` whenever needed. It still allows us to have conditionals and it feels more consistent with `defaultProps`. This is what I mean:

```js
constructor(props) {
  super(props)

  this.state = this.defaultState = {
    foo: 'bar',
    bar: props.baz ? 'yep' : 'nope'
  }

  // Or, if you don't want to chain assignments:
  this.defaultState = {
    foo: 'bar',
    bar: props.baz ? 'yep' : 'nope'
  }

  this.state = this.defaultState
}

someLifecycleMethod() {
  this.setState(this.defaultState)
}
```

If you still want to make it a method, I'd call it `getDefaultLocalState()` as function names should have a verb.

frontend/containers/DocumentList/DocumentList.jsx
```diff
@@ -87,12 +87,14 @@ class DocumentList extends Component {
     const historyKeyMatch = pathKey === previousPathKey
     const apisWithoutCollections = state.api.apis.filter(api => !api.hasCollections).length
 
+    if(prevProps.collection !== this.props.collection || prevProps.page !== this.props.page) {
```
eduardoboucas: Missing space after `if`. (How come the linter didn't complain?)

mingard: Not sure

frontend/containers/DocumentList/DocumentList.jsx
mingard: I see your point, however I believe it makes sense to use a method, as the default state has real potential to become pretty large and contain complex conditional parameters. This is for consistency and readability. 

Method name changed.

eduardoboucas: State within components should always remain thin and lightweight, as the heavy-lifting should always belong in the store. If we find ourselves with pretty large state at the component level, we're probably doing something wrong.

mingard: Agreed, but container level is bound to have a number of local state values, as well as logic behind setting them.

eduardoboucas: The logic for containers is the same, they should still be pretty lightweight in local state. Not sure what you mean by the logic behind setting that state, but I still think the best place to initialise the state is within the bounds of the constructor block, as it's common practice and expected behaviour. Whatever you prefer, though.

mingard: I'm still unconvinced. Will review this post-mvp.

PR #194, opened 2017-04-25
Replace username with email

eduardoboucas:
This PR replaces `username` with `email` on user data. It should be merged with [this PR](https://github.com/dadiplus/publish-api/pull/6) on the API repository, which updates the password hook.


eduardoboucas:
@mingard I think I might as well include in this PR the changes to move the sign in to an action, so maybe wait for that before merging?

mingard:
@eduardoboucas cool. And we need to remove the requirement for username from API

app/lib/controllers/session.js
```diff
@@ -5,14 +5,14 @@ const Api = require(`${paths.lib.models}/api`)
 
 const Session = function () {}
 
-Session.prototype.authorise = function (username, password, next) {
```
mingard: @eduardoboucas has this been tested? I seem to recall an issue with changing this parameters name, and that restify wouldn't allow is as an auth middleware routine

eduardoboucas: It has been tested. I left the argument of the `authorise` function as `username` precisely for that reason. Took me a while to figure out what was breaking though!

mingard: Ah that was where it was. We should document that

eduardoboucas: I think I'll add a comment in the code for now, just to make that clear?

mingard: Yes please 🥇

eduardoboucas:
@mingard This is ready for review.

PR #197, opened 2017-04-25
Feat/save commands

mingard:


frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
```diff
@@ -74,9 +75,14 @@ class DocumentEditToolbar extends Component {
   constructor(props) {
     super(props)
 
+    this.keyboard = new Keyboard()
     this.onSave = null
   }
 
+  componentDidMount() {
+    this.keyboard.on('cmd+s').do(this.saveDocument.bind(this))
```
eduardoboucas: I would do:

```
this.keyboard.on('cmd+s').do(this.handleSave.bind(this, 'save'))
```

Because:

- `saveDocument` is supposed to be called by `handleSave`, not the other way around
- Removes the need for the "Apply default" block you've added to `saveDocument`
- Makes it more explicit which save mode is attached to Cmd+S

frontend/lib/keyboard.js
```diff
@@ -155,7 +157,8 @@ export class Keyboard extends Keys {
 
   keydown (event) {
     if (event.keyCode) {
-      if (this.findShortcut(event.keyCode)) {
+      let hasNext = this.findShortcut(event.keyCode)
```
eduardoboucas: This could be a `const`.

frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
mingard: I tried that, but it didn't work

eduardoboucas: If you give a bit more detail into what happened, I can try to help.

mingard: Nothing, it just doesn't save

eduardoboucas: Okay, mind if I have a go?

mingard: Sure!

eduardoboucas: I've just changed to what I had proposed and it worked fine for me. Could you pull and test, please?

mingard: Worked for me!

PR #198, opened 2017-04-25
feat: style file drag state closes #150

### LINKED ISSUE #150: Design for drag-and-drop over state
UI required in support of the drag-and-drop asset upload method.
screen shot 2017-04-03 at 13 25 13

mingard:


frontend/components/FileUpload/FileUpload.css
```diff
@@ -16,4 +16,22 @@
...
+  padding: 10px;
+  border: 1px solid transparent;
+  border-radius: 2px;
+}
+
+.dropzone-active {
+    border: 1px solid var(--theme-colour-data, #4A91FF);
+    background-color: #e3eeff;
+    box-shadow: 0 0 14px rgba(0,0,0,.1) inset;
+    -webkit-transition: box-shadow 300ms linear;
```
eduardoboucas: We don't need to add browser prefixes, as that's automatically done by PostCSS.

Also, you have 3 `transition` declarations applied on the same selector. The last one will override the other two. 

Don't you mean:

```css
transition: box-shadow 300ms linear, background-color 300ms linear, border 300ms linear;
```

mingard: @eduardoboucas ah, indeed we do. There was no override, but that's a cleaner declaration

Issue #195, opened 2017-04-25
Reset password view


eduardoboucas:
This is the only view left to do. From what I can see, we need:

**Back-end:**

- A password reset endpoint
- Generate a reset password token
- Send an email with a link to the endpoint + token
- Verify the token and reset the password

**Front-end:**

- The styled view

Thoughts @mingard?

eduardoboucas:
Okay. If you let me know how the Publish endpoint for requesting a password reset will look like, I can work on the front-end part.

eduardoboucas:
Do you have any information on what the endpoint will look like? Could we define it as `POST /reset` receiving an `email` property with the email address, returning a 200 with `success` as `true` or `false` depending on whether or not the address is valid, and a 500 if the email couldn't be sent?

mingard:
@eduardoboucas that sounds reasonable. We could use `dadi/queue` for email services, and have API handle this, not Publish.

eduardoboucas:
Isn't it a bit overkill to have a requirement for Queue just so we can send password reset emails, which will likely have a very low volume? Also, are you still happy to work on the back-end logic for this, or do you need help?

mingard:
It would allow us to retain pending tasks and handle resending if there is a failure so better for error handling and non-blocking.

eduardoboucas:
I understand that, but from what I understand, Queue is a microservice on its own, so at it currently stands it means deploying another service/web server. Is it reasonable to ask people to do this with the justification of being able to send password reset emails?

mingard:
As per discussion, the following are required to complete this issue:
https://github.com/dadiplus/publish-api/issues/8
https://github.com/dadi/api/issues/254
https://github.com/dadi/publish/issues/210
https://github.com/dadi/publish/issues/211

PR #204, opened 2017-04-28
Password reset view + View component

eduardoboucas:
This PR adds the password reset view. The call-to-action isn't hooked up to an action yet, as I'm waiting for feedback on #195.

In addition to that, it addresses an issue we currently have with the `Router` component in the `App` container. Looking at the existing routes we see a `authenticate` prop added to some routes, which makes you think it establishes which routes need to be under authentication. This isn't the case though, and this prop isn't doing anything (you can remove it and observe that the app keeps working in the exact same way).

What's currently taking care of ensuring authentication is in place for most routes is [this check](https://github.com/dadi/publish/blob/master/frontend/containers/App/App.jsx#L82-L84), which checks for a user and redirects to `/sign-in` when none is found. This is fine if `/sign-in` is the only route that is unauthenticated, but with the introduction of the password reset view it becomes an issue.

To get around that, I created a thin `View` component that receives as prop the component that should be rendered for a particular route, and this time we actually look for the `authenticate` prop — if it exists and the user isn't signed in, we take care of the redirect.


eduardoboucas:
We could keep the current structure and listen for the `authenticate` prop in each view, but this would mean repetition of the logic for checking+redirecting on every single view, and it would also mean that every view would need to have access to the store, which isn't currently the case.

That's why I added the `View` wrapper.

PR #206, opened 2017-04-28
Fix/document list fix

mingard:
This PR fixes an issue with document list view failing to render when navigating from index. This was caused by migrating to latest Preact and removing error provoking routing checks within the `<Router>` component.

It also introduces a global `onChange` event for routes, which in turn handles authentication redirects more elegantly than previous iterations of the same method.

### DIFF
--- a/frontend/containers/App/App.jsx
+++ b/frontend/containers/App/App.jsx
@@ -108,6 +108,22 @@ class App extends Component {
     // TO-TO Handle leaving a room when the component changes
   }
 
+  handleRouteChange(event) {
+    const {state} = this.props
+    const {authenticate} = event.current.attributes
+    const {user} = state
+    const notAuthenticated = user.status === Constants.STATUS_FAILED
+    if (authenticate && notAuthenticated) {
+       route('/sign-in')
+
+       return null
+     }
+ 
+     if (user.status === Constants.STATUS_LOADING) {
+       return null
+     }
+  }
+
   render() {
     const {history, state} = this.props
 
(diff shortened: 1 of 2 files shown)


frontend/containers/DocumentList/DocumentList.jsx
```diff
@@ -184,6 +176,22 @@ class DocumentList extends Component {
     return this.renderDocumentList()
   }
 
+  componentWillMount() {
+    this.checkStatusAndLoad()
+  }
+
+  checkStatusAndLoad() {
```
eduardoboucas: Super minor thing: I'd either name this `checkStatusAndFetch` or rename `fetchDocuments` to `loadDocuments`, just so that it's clear that one thing triggers the other.

frontend/containers/App/App.jsx
```diff
@@ -108,6 +108,22 @@ class App extends Component {
     // TO-TO Handle leaving a room when the component changes
   }
 
+  handleRouteChange(event) {
+    const {state} = this.props
+    const {authenticate} = event.current.attributes
+    const {user} = state
+    const notAuthenticated = user.status === Constants.STATUS_FAILED
+    if (authenticate && notAuthenticated) {
```
eduardoboucas: Could you add a line break after the declarations?

frontend/containers/DocumentList/DocumentList.jsx
mingard: Resolved

Issue #214, opened 2017-05-10
Support for multiple and single reference field data


mingard:
We have a flag for multiple, which is used in Publish and exists only in the `publish` block.

Apart from changing the way content is displayed in Publish, this flag decides whether the Reference field can have one or many entries. 

The requirements outstanding, and for discussion:

- Multiple defaults to true
- If set to false, store as a single ObjectId, no array
- If true, store as an array of ObjectIds
- If limit is set, always store as an array, even if limit is 1
- Limit 0 defaults to no limit

 @jimlambie we may want to consider moving this out of the `publish` block, if we believe this should be enforced in API too

eduardoboucas:
I can get started on this. If we decide to make it part of API, I'll just need to look for `settings.limit` instead of `publish.limit`.

mingard:
Yes please

eduardoboucas:
I've made API format multiple media documents (in array format as well as single ObjectID). It doesn't seem to currently resolve non-media references in array format, so I'll address that first before I can finish the Publish piece. 👷

jimlambie:
@mingard "multiple" should already be in the Reference field settings, not the publish block?

mingard:
@eduardoboucas sounds good, thanks!
@jimlambie Okay, what affect does if have? Is there a limit value that limits the number of referenced documents in the array?

eduardoboucas:
Done in #232.

PR #232, opened 2017-05-12
Add support for multi-document reference fields

eduardoboucas:
This PR implements #214, adding support for reference fields with multiple documents, based on the `settings.limit` property.

This is for non-media reference fields only — do we want to do the same for media fields, @mingard?

### DIFF
--- a/frontend/components/FieldImage/FieldImageEdit.jsx
+++ b/frontend/components/FieldImage/FieldImageEdit.jsx
@@ -106,20 +106,25 @@ export default class FieldImageEdit extends Component {
       value
     } = this.props
     const displayName = schema.label || schema._id
-    const src = this.getImageSrc(value)
-    const isReference = schema.type === 'Reference'
     const fieldLocalType = schema.publish && schema.publish.subType ? schema.publish.subType : schema.type
     const href = buildUrl(...onBuildBaseUrl(), 'select', schema._id)
+    const isReference = schema.type === 'Reference'
+    const singleFile = schema.settings && schema.settings.limit === 1
+    const values = value && !(value instanceof Array) ? [value] : value
 
     return (
       <Label label={displayName}>
-        {value &&
+        {values &&
           <div class={styles['value-container']}>
-            <img
-              class={styles.preview}
-              src={src}
-            />
-            
+            <div class={styles.thumbnails}>
+              {values.map(value => (
+                <img
+                  class={styles.thumbnail}
+                  src={this.getImageSrc(value)}
+                />
+              ))}
+            </div>
+
             <Button
               accent="destruct"
               size="small"
@@ -129,7 +134,7 @@ export default class FieldImageEdit extends Component {
           </div>          
         }
 
-        {!value &&
+        {!values &&
           <div>
             <div class={styles.placeholder}>
               <Button
@@ -144,6 +149,7 @@ export default class FieldImageEdit extends Component {
               <FileUpload
                 allowDrop={true}
                 accept={config.accept}
+                multiple={!singleFile}
                 onChange={this.handleFileChange.bind(this)}
               />
             </div>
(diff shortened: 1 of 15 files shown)


mingard:
@eduardoboucas yes please - all Reference fields

eduardoboucas:
Sure thing! It'll require some more work (e.g. requesting multiple signed URLs, handling multiple uploads, etc.), so once you're happy with this PR I suggest we merge it to avoid keeping `master` outdated for too long.

frontend/containers/DocumentListToolbar/DocumentListToolbar.jsx
```diff
@@ -264,13 +266,14 @@ class DocumentListToolbar extends Component {
...
     // documents. For now, we just get the first selected document.
-    const selectedDocumentId = state.documents.selected[0]
-    const selectedDocument = documentsList.find(document => {
-      return document._id === selectedDocumentId
-    })
+    const selectedDocuments = state.documents.selected.map(documentId => {
+      return documentsList.find(document => {
+        return document._id === documentId
+      })
+    }).filter(Boolean)
```
mingard: @eduardoboucas shorthand suggestion ;)
```javascript
const selectedDocuments = state.documents.selected
  .map(documentId => documentsList.find(document => document._id === documentId))
  .filter(Boolean)
```

frontend/containers/DocumentList/DocumentList.jsx
```diff
@@ -417,9 +421,11 @@ class DocumentList extends Component {
   }
 
   renderField(fieldName, schema, value) {
-    const fieldType = schema.publish && schema.publish.subType ? schema.publish.subType : schema.type
+    const fieldType = schema.publish && schema.publish.subType ?
```
mingard: I think I wrote this originally, but could we wrap multiple conditions perhaps?
```javascript
(schema.publish && schema.publish.subType)
```

frontend/components/FieldImage/FieldImageEdit.jsx
```diff
@@ -174,26 +180,40 @@ export default class FieldImageEdit extends Component {
...
-    const reader = new FileReader()
-    
-    reader.onload = event => {
-      if (typeof onChange === 'function') {
-        onChange.call(this, schema._id, {
+    const singleFile = schema.settings && schema.settings.limit === 1
+
+    let processedFiles = []
+
+    for (let index = 0; index < files.length; index++) {
```
mingard: Just curious - any reason for not using this?
```javascript
files.forEach()
```

frontend/components/FieldImage/FieldImageEdit.jsx
```diff
@@ -106,20 +106,25 @@ export default class FieldImageEdit extends Component {
       value
     } = this.props
     const displayName = schema.label || schema._id
-    const src = this.getImageSrc(value)
-    const isReference = schema.type === 'Reference'
     const fieldLocalType = schema.publish && schema.publish.subType ? schema.publish.subType : schema.type
     const href = buildUrl(...onBuildBaseUrl(), 'select', schema._id)
+    const isReference = schema.type === 'Reference'
+    const singleFile = schema.settings && schema.settings.limit === 1
+    const values = value && !(value instanceof Array) ? [value] : value
```
mingard: Suggestion:
```javascript
const values = (value && Array.isArray(value)) ? value : [value]
```

frontend/containers/DocumentListToolbar/DocumentListToolbar.jsx
eduardoboucas: I personally like to use shorthand arrow functions for simple and short statements. IMO having a `.find()` inside a shorthand `.map()` is unnecessarily cryptic.

frontend/containers/DocumentList/DocumentList.jsx
eduardoboucas: Sure, will change that.

frontend/containers/DocumentListToolbar/DocumentListToolbar.jsx
mingard: You could format it in another order. I believe it's more readable, but not a styleguide issue so up to you.

frontend/components/FieldImage/FieldImageEdit.jsx
eduardoboucas: `files` is an instance of [FileList](https://developer.mozilla.org/en/docs/Web/API/FileList), not Array, so it doesn't have a `forEach` method.

mingard: ES6: `const files = [...FileList]`

frontend/components/FieldImage/FieldImageEdit.jsx
eduardoboucas: If `value` is `null`, `values` would end up as `[null]` which evaluates to `true`.

mingard:
```javascript
const values = (value && !Array.isArray(value)) ? [value] : value
```

mingard: Reason for suggestion is that `isArray` is considered more robust.

eduardoboucas: Sure. I think we're using `instanceof` in a few places, will need to change those too.

mingard: Sorry to make more work for you!

eduardoboucas: No worries, I'll change shortly.

eduardoboucas: @mingard Done. Changed all instances I could find.

eduardoboucas:
@mingard can we merge this?

PR #238, opened 2017-05-17
feat: refractor filters

mingard:
feat: filters are now updated on return and on apply click
feat: filter inputs are updated on keyup

### DIFF
--- a/frontend/components/DocumentFilter/DocumentFilter.css
+++ b/frontend/components/DocumentFilter/DocumentFilter.css
@@ -2,28 +2,33 @@
   display: flex;
   flex-wrap: wrap;
   width: 100%;
-  margin-bottom: 10px;
+  justify-content: space-between;
+  align-items: center;
+  justify-content: center;
 }
 
 .select {
-  width: 100%;
+  flex-grow: 50;
   border: 0;
   height: 25px;
   line-height: 25px;
   font-size: 13px;
   background-color: #ffffff;
   -webkit-appearance: menulist-button;
+  margin-left: 5px;
 }
 
 .input {
-  width: 100%;
+  width: 50%;
   border-radius: 4px;
   border: 1px solid #e4e4e4;
-  width: 100%;
   height: 25px;
   line-height: 25px;
   font-size: 13px;
-  padding: 5px;
+  /*padding: 5px;*/
+  display: flex;
+  margin: 10px 0;
+  justify-content: space-between;
 }
 
 input:focus {
@@ -35,18 +40,26 @@ input:focus {
   width: 25px;
 }
 
-.control + .control {
-  margin-left: 5px;
-}
-
 .control-field {
-  flex-grow: 50;
+  flex-grow: 30;
 }
 
 .control-analyser {
+  margin: 10px;
   flex-grow: 20;
 }
 
 .control-value {
-  flex-grow: 100;
+  flex-grow: 50;
+  margin-right: 10px;
+}
+
+.control-value-select {
+  
+}
+
+.filter-container {
+  flex-grow: 70;
+  display: flex;
+  justify-content: space-between;
 }
--- a/frontend/components/DocumentFilter/DocumentFilter.css
+++ b/frontend/components/DocumentFilter/DocumentFilter.css
@@ -53,10 +53,6 @@ input:focus {
   margin-right: 10px;
 }
 
-.control-value-select {
-
-}
-
 .filter-container {
   flex-grow: 70;
   display: flex;
--- a/frontend/components/DocumentFilter/DocumentFilter.css
+++ b/frontend/components/DocumentFilter/DocumentFilter.css
@@ -25,7 +25,6 @@
   height: 25px;
   line-height: 25px;
   font-size: 13px;
-  /*padding: 5px;*/
   display: flex;
   margin: 10px 0;
   justify-content: space-between;
@@ -55,7 +54,7 @@ input:focus {
 }
 
 .control-value-select {
-  
+
 }
 
 .filter-container {
--- a/frontend/components/DocumentFilter/DocumentFilter.jsx
+++ b/frontend/components/DocumentFilter/DocumentFilter.jsx
@@ -26,6 +26,11 @@ export default class DocumentFilter extends Component {
     fields: proptypes.object,
 
     /**
+     * An object containing the current applied filters.
+     */
+    filters: proptypes.object,
+
+    /**
      * The index of the current filter within a list of filters.
      */
     index: proptypes.number,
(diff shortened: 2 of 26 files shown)


frontend/components/DocumentFilter/DocumentFilter.css
```diff
@@ -35,18 +39,26 @@ input:focus {
...
   flex-grow: 20;
 }
 
 .control-value {
-  flex-grow: 100;
+  flex-grow: 50;
+  margin-right: 10px;
+}
+
+.control-value-select {
```
eduardoboucas: Anything missing here?

frontend/components/DocumentFilter/DocumentFilter.jsx
```diff
@@ -60,80 +70,60 @@ export default class DocumentFilter extends Component {
...
+              key={key}
+              value={key}
+            >
+              {fields[key].label}
+            </option>
+          ))}
+        </select>
+
+        {(field && type) && (
+          <FieldFilter
```
eduardoboucas: Props should be in alphabetical order.

frontend/components/DocumentFilters/DocumentFilters.jsx
```diff
@@ -37,13 +42,17 @@ export default class DocumentFilters extends Component {
 
     const {collection} = this.props
     const paramFilters = this.getFiltersFromParams()
-    this.state.filters = paramFilters
+
+    this.state = {
```
eduardoboucas: Keys should be in alphabetical order. The linter should be picking up this one.

frontend/components/DocumentFilters/DocumentFilters.jsx
```diff
@@ -37,13 +42,17 @@ export default class DocumentFilters extends Component {
...
+      filters: paramFilters,
+      dirty: false
+    }
   }
 
   getFiltersFromParams () {
     const {filters} = this.props
     return (filters ? objectToArray(filters, 'field') : [])
```
eduardoboucas: Empty line before return.

frontend/components/DocumentFilters/DocumentFilters.jsx
```diff
@@ -129,8 +155,8 @@ export default class DocumentFilters extends Component {
     Object.assign(newFilters[index], filter)
     this.setState({filters: newFilters})
 
-    if (filter.value) {
-      this.updateUrl()
+    if (filter.value !== undefined) {
```
eduardoboucas: Shouldn't this be `typeof filter.value !== 'undefined'`?

frontend/components/DocumentFilters/DocumentFilters.jsx
```diff
@@ -140,9 +166,8 @@ export default class DocumentFilters extends Component {
     const {updateUrlParams} = this.props
 
     let validFilters = filters.filter(entry => {
-      let isValid = Object.keys(entry).filter(key => {
-        return Object.is(entry[key], null)
-      })
+      let isValid = Object.keys(entry)
+        .filter(key => Object.is(entry[key], null))
```
eduardoboucas: Do we need `Object.is()` here? It's not supported by any version of IE and it seems that, in this case, we can just check if the value is falsy or compare against `null`.

frontend/components/FieldDateTime/FieldDateTimeFilter.jsx
```diff
@@ -0,0 +1,120 @@
...
+              <option
+                selected={type === key}
+                key={key}
+                value={key}
+              >
+                {this.filterTypes[key]}
+              </option>
+            ))}
+          </select>
+            <TextInput
```
eduardoboucas: Props could be in alphabetical order.

frontend/components/FieldString/FieldStringFilter.jsx
```diff
@@ -0,0 +1,103 @@
...
+              <option
+                selected={type === key}
+                key={key}
+                value={key}
+              >
+                {this.filterTypes[key]}
+              </option>
+            ))}
+          </select>
+            <TextInput
```
eduardoboucas: Props could be in alphabetical order.

frontend/components/DocumentFilters/DocumentFilters.jsx
mingard: @eduardoboucas it doesn't seem to. Confirm that this is picked up with you run `npm run test`?

frontend/components/DocumentFilters/DocumentFilters.jsx
mingard: @eduardoboucas ES6 supports this

eduardoboucas: It's just for consistency, as we've been using `typeof` everywhere else.

mingard: We should change that. `=== undefined` has always worked, but previously you could globally defined `undefined` making it a useless exact compare. I'll go through and resolve this in other places if you like. Just raise a ticket.

frontend/components/DocumentFilters/DocumentFilters.jsx
mingard: Yes, it could be `.filter(Boolean)`

frontend/components/DocumentFilter/DocumentFilter.css
mingard: What do you mean?

eduardoboucas: Any reason to include an empty selector?

frontend/components/DocumentFilters/DocumentFilters.jsx
eduardoboucas: I see we're using `Object.is` [here](https://github.com/dadi/publish/blob/master/frontend/middleware/router.js#L57) as well, which will cause trouble in IE. We should probably revisit that.

frontend/components/DocumentFilter/DocumentFilter.css
mingard: No, should have been removed

frontend/components/DocumentFilters/DocumentFilters.jsx
mingard: @eduardoboucas That depends on whether we'll ever support IE. Edge has full support

frontend/components/DocumentFilters/DocumentFilters.jsx
mingard: All updated

frontend/components/DocumentFilters/DocumentFilters.jsx
eduardoboucas: Shouldn't that decision be made when the cost of supporting IE has an impact on efficiency/maintainability? I see no reason to use `Object.is` in this case or in the case I linked to, which should be a simple comparison.

mingard: In the case you linked to, the alternative is to check that an object is equal to one or another value. This check is what `Object.js` is designed for. 

You may have overlooked this, but there are many methods we're using that aren't supported by IE.

frontend/components/DocumentFilter/DocumentFilter.jsx
mingard: Not showing up as ordered because the comment is on the top line

frontend/components/DocumentFilters/DocumentFilters.jsx
mingard: Done

PR #251, opened 2017-06-06
feat: start backend tests

mingard:


test/unit/config.test.js
```diff
@@ -0,0 +1,102 @@
...
+        port: true,
+        database: true,
+        version: true,
+        menu: true
+      } 
+    }
+    it ('should return object', () => {
+      expect(config.getFrontendProps()).toBeInstanceOf(Object)
+    })
+    const schema = config.getFrontendProps(input, {})
```
eduardoboucas: We shouldn't have any logic running outside the test functions. Could this be moved inside the `.it()`?

test/unit/config.test.js
```diff
@@ -0,0 +1,102 @@
+const globals = require(`${__dirname}/../../app/globals`) // Always required
+const config = require(`${__dirname}/../../app/config`)
+
+describe('Config', () => {
+  it ('should export object', () => {
+    expect(config).toBeInstanceOf(Object)
+  })
+  describe('#getFrontendProps()', () => {
```
eduardoboucas: Could you add a line break after each test function, as seen in other products?

test/unit/config.test.js
```diff
@@ -0,0 +1,102 @@
...
+        database: true,
+        version: true,
+        menu: true
+      } 
+    }
+    it ('should return object', () => {
+      expect(config.getFrontendProps()).toBeInstanceOf(Object)
+    })
+    const schema = config.getFrontendProps(input, {})
+    it ('should return only input props with true "availableInFrontend" key', () => {
```
eduardoboucas: Minor thing, but we typically do `it()` and not `it ()` (i.e. no space).

test/unit/config.test.js
mingard: The results are reused in a number of tests after

test/unit/config.test.js
mingard: Done

test/unit/config.test.js
eduardoboucas: Test shouldn't create state that is used by other tests. If you need to reuse the result of an operation in multiple tests, you should either re-create that state in every single test or have a global variable that is updated on a `before`/`after` hook, that runs before of after each test.

A common pattern is to use a `before` to prepare the state for a test and an `after` to reset/cleanup, so that the next test to run has a blank canvas.

eduardoboucas: Example: https://github.com/dadi/publish/blob/master/frontend/components/Button/Button.test.jsx#L11-L19

mingard: It is creating an Object which is then analysed 3x ways. It seems counterintuitive to recreate the object if the entire point of the 3 tests is to check the same subject.

mingard: Jest documentation seems to suggest this is okay too:
```
describe('Beware of a misunderstanding! A sequence of dice rolls', () => {
  const expected = [1, 2, 3, 4, 5, 6];
  it('matches even with an unexpected number 7', () => {
    expect([4, 1, 6, 7, 3, 5, 2, 5, 4, 6])
      .toEqual(expect.arrayContaining(expected));
  });
  it('does not match without an expected number 2', () => {
    expect([4, 1, 6, 7, 3, 5, 7, 5, 4, 6])
      .not.toEqual(expect.arrayContaining(expected));
  });
});
```
source: https://facebook.github.io/jest/docs/expect.html#expectarraycontainingarray

eduardoboucas: I'm just advising on best practices that we use in our test suites. Re-using an object in several tests is a very common pattern and `before`/`after` hooks are what we typically use.

Other examples:

- api-wrapper-core: https://github.com/dadi/api-wrapper-core/blob/master/test/unit/terminators.js#L61-L69

- API: https://github.com/dadi/api/blob/d8921e678ec427153e2f5d48d59897523e3c6cd1/test/unit/cache.js#L31-L42

- Web: https://github.com/dadi/web/blob/6c60c1fbc9604e53de6fed19df8de17d68f65e5f/test/unit/session.js#L35-L41

In any case, do as you prefer.

mingard: In the Web example I can see justification in starting and stopping a mock server for each test. In the API example, I can see the justification in rebuilding cache. The difference is mutation and/or environmental changes to the reused variables. In this case we're talking about an immutable object being analysed for 3 different reasons. The idea of recreating the object is itself a worse test as the checks explicitly review the same subject.

PR #258, opened 2017-06-13
feat: background image random svg

mingard:


frontend/components/Main/Main.jsx
```diff
@@ -11,6 +11,10 @@ import NotificationCentre from 'containers/NotificationCentre/NotificationCentre
  * The main content body element.
  */
 export default class Main extends Component {
+  componentDidMount() {
```
eduardoboucas: Could this be `componentWillMount()` to load the background as early as possible?

mingard: Ah yes

PR #264, opened 2017-06-14
Feat/post install

mingard:
This feature addresses the requirement for an automated postinstall build event for npm module use.


index.js
```diff
@@ -15,12 +15,10 @@ Publish.prototype.start = function (options) {
 // Run-type switch
 if (require.main === module) {
   // App called directly
-  module.exports = new Publish().start()
+  module.exports = new Publish().run(true)
```
eduardoboucas: Looks great! The only thing I'd probably change, and that is down to personal taste, is change the `Publish` constructor to accept an options object, rather than individual arguments. This would make the call to it a lot more descriptive.

```js
// Before
module.exports = new Publish().run(true)

// After
module.exports = new Publish().run({isStandalone: true})
```

If the constructor grows to accept more arguments, it helps to keep it short and clean.

mingard: And then change
```js
App.prototype.start = function (isStandalone) {
  if (!isStandalone) {
    const watcher = new Watch().start()

    // Overwrite default config with app-specific module
    config = watcher.getConfig(config)
  }
  this.server = new Server()
  return this.server.start()
}
```
to
```js
App.prototype.start = function ({isStandalone = false}) {
  if (!arguments.isStandalone) {
    const watcher = new Watch().start()

    // Overwrite default config with app-specific module
    config = watcher.getConfig(config)
  }
  this.server = new Server()
  return this.server.start()
}
```
right?

mingard: @eduardoboucas better?

eduardoboucas: 👍

PR #267, opened 2017-06-19
feat: check, create/update required hooks

mingard:


eduardoboucas:
Looks good! Just a question. I had to make changes to the auth hook as part of the fix to the missing sign in errors. Going forward, how should we manage updates to this hook, since its source will be both on this repo and on the Publish API repo?

mingard:
@eduardoboucas It would be useful to be able to pull them directly from the publish-api repo, however it is closed. For now, I'd commit any hook or collection change to publish-api repo, then update the version stored in `scripts/api-install/hooks` and `scripts/api-install/collections` respectively.

PR #270, opened 2017-06-20
Feat/helper tests

mingard:
Testing url regex against valid and invalid string values

### DIFF
--- a/test/unit/lib/helpers/regex.test.js
+++ b/test/unit/lib/helpers/regex.test.js
@@ -9,12 +9,15 @@ describe('Regex: URL', () => {
   describe('URL test http', () => {
     const http = 'http://example.com'
     const match = regex.url.exec(http)
+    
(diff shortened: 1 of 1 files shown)


test/unit/lib/helpers/regex.test.js
```diff
@@ -0,0 +1,59 @@
...
+    })
+    it (`Should have a first argument of 'http'`, () => {
+      expect(match[1]).toBe('http')
+    })
+  })
+
+  describe('URL test https', () => {
+    const https = 'https://example.com'
+    const match = regex.url.exec(https)
+    it (`Should not return null for '${https}'`, () => {
```
eduardoboucas: Looks good! The only thing I'd change is add a line break between each test, so that it's consistent with our other test suites. And also remote the space on the `it()` call.

mingard: I did it again didn't I?

PR #281, opened 2017-06-29
Feat/ssl middleware

mingard:


app/lib/server/index.js
```diff
@@ -25,74 +25,118 @@ const getApisBlockWithUUIDs = apis => {
...
 }
 
 /**
  * Start server
  * @return {Promise}  Add Listener
  */
 Server.prototype.start = function () {
-  return new Promise((resolve, reject) => {
+  let listenerQueue = []
+  return new Promise(resolve => {
```
eduardoboucas: I don't think we need to create a new Promise here, you could just return the result of `Promise.all`, no?

mingard: @eduardoboucas this is true - its an artefact from a previous iteration that had a single promise

PR #292, opened 2017-07-25
chore: change package reference

mingard:
Introduces a server restart method required for https://github.com/dadi/ssl/pull/3


mingard:
@jimlambie thanks! Let me know when SSL is npm published and I'll merge this.

PR #302, opened 2017-08-09
Feat/route based urls

mingard:


app/lib/helpers/collection-routes.js
```diff
@@ -0,0 +1,158 @@
...
+  return Math.max(...this.fieldReferenceDepth(collections, fields, depth))
+}
+
+CollectionRoutes.prototype.fieldReferenceDepth = function (collections, fields, depth) {
+  return fields
+    .map(field => {
+      const colName = field.settings.collection
+      const subCol = this.getCollectionBySlug(collections, colName)
+
+      // If the collection doesn't exist, or it is.
```
eduardoboucas: Or it is? 😄

app/lib/models/collection.js
```diff
@@ -0,0 +1,42 @@
...
+  // new Api(config)
+  return this.getCollections()
+    .then(apiCollections => new CollectionRoutes(apiCollections))
+}
+
+Collection.prototype.getCollections = function () {
+  return Promise.all(
+    config
+      .get('apis')
+      .map(api => new Api(api)
```
eduardoboucas: Personal preference: whenever the content of the arrow function is more complex than a single value, I'd use a normal function body with a `return`, otherwise the indentation is harder to understand. For example:

```js
Collection.prototype.getCollections = function () {
  return Promise.all(
    config
      .get('apis')
      .map(api => {
        return new Api(api)
          .getCollections()
          .then(res => this.getSchemas(api, res.collections))      
      })
  )
}
```

app/lib/models/collection.js
```diff
@@ -0,0 +1,42 @@
...
+        .getCollections()
+        .then(res => this.getSchemas(api, res.collections))
+      )
+  )
+}
+
+Collection.prototype.getSchemas = function (api, collections) {
+  return Promise.all(
+    collections
+      .map(collection => new Api(api)
```
eduardoboucas: (Same as above)

frontend/containers/DocumentEdit/DocumentEdit.jsx
```diff
@@ -365,6 +371,30 @@ class DocumentEdit extends Component {
...
+      onBuildBaseUrl,
+      onGetRoutes,
+      state
+    } = this.props
+    const sectionUrlBase = onBuildBaseUrl()
+
+    if (collection === Constants.AUTH_COLLECTION) {
+      return buildUrl(...sectionUrlBase, collectionSection.slug)
+    }
+    if (method === 'new') {
```
eduardoboucas: An empty line between these `if` blocks would be 💣

app/lib/models/collection.js
mingard: @eduardoboucas I've been in two minds about that, but tend to agree. Will change this.

app/lib/helpers/collection-routes.js
mingard: Ha! No idea. Will try and work out what I was saying.

PR #308, opened 2017-08-17
Feat/field reduce

mingard:
In this PR:
- Reduce the number of fields loaded in list view by filtering by `publish.display.list` with a fallback default of true.
- Adds jest watch script to package.json.


mingard:
@eduardoboucas Sound logic.

PR #327, opened 2017-10-10
feat: handle api connection error with message

mingard:
Outputs an error when API returns 401 or 404.


frontend/components/ErrorMessage/ErrorMessage.jsx
```diff
@@ -78,10 +78,22 @@ export default class ErrorMessage extends Component {
           title: 'API failure'
         }
 
+      case Constants.API_CONNECTION_ERROR:
```
eduardoboucas: Add a test for this case? 😇

mingard: Done!

Issue #343, opened 2017-10-16
Unselected select boxes don't use the placeholder as the selected option


jimlambie:
I'm not sure how to explain it:

<img width="178" alt="screen shot 2017-10-16 at 14 59 54" src="https://user-images.githubusercontent.com/116263/31615877-f4d78c4c-b282-11e7-9aa4-d3c7fef50db7.png">
<img width="269" alt="screen shot 2017-10-16 at 15 00 00" src="https://user-images.githubusercontent.com/116263/31615890-f9d96f26-b282-11e7-8a34-428400103dee.png">

Browser: Chrome

mingard:
@jimlambie An out of focus select node doesn't display the selected value?

jimlambie:
Correct. Or - are you giving me the explanation?

mingard:
No, just making sure we're on the same page. I'll look into this.

jimlambie:
@mingard it's not really a selected value. It's a select box with no selected options, that looks as if it has a placeholder which should be displayed

mingard:
@jimlambie it was just a missing value on the default option.

PR #344, opened 2017-10-16
Fix/select label with fallback

mingard:
This PR 
- adds a placeholder value to the select box
- tests for label fallback
- Defaults to the field id 

fixes https://github.com/dadi/publish/issues/343


mingard:
@jimlambie for some reason, the default `<option>` needed a value for the placeholder.

Issue #396, opened 2018-06-12
Implement API ACL


danwdart:
Could use @dadi/api-wrapper through an isomorphic fetch.

eduardoboucas:
Could you expand a bit on this one?

Not sure if this is what you're referring to, but we currently don't have server-side rendering for the UI components. It's something we discussed, but ultimately we felt that the potential performance gains weren't important enough to justify the added complexity.

We use API wrapper extensively in Publish to form any API queries ([`@dadi/api-wrapper-core`](https://www.npmjs.com/package/@dadi/api-wrapper-core), to be more precise, which is the non-Node version of API wrapper). My thinking was that instead of making the Publish server relay requests to API, the front-end could use `@dadi/api-wrapper-core` directly to get data from API, setting the [callback function](https://github.com/dadi/api-wrapper-core#defining-a-callback) to use [fetch](https://github.com/github/fetch) or any other Ajax library.

Does this align with what you're thinking?

danwdart:
I understand the first point.
I see what you mean about the second point - so that we would just need to change the destination and add the auth headers - seeing as we're already using api-wrapper-core.

danwdart:
Yes, I believe it's this one - https://github.com/dadi/publish/blob/master/frontend/lib/api-bridge-client.js - it should hit the API itself this time.

danwdart:
This would probably require exposing an access token to the frontend code - if not a whole new pair of credentials - will try out this working before ACL

eduardoboucas:
The idea with the new ACL piece is that each Publish user will have their own client ID+secret pair in API, so Publish will be able to request a bearer token on their behalf (by pinging the `/token` endpoint in API). The resulting bearer token will then be used by the front-end application to communicate with API directly.

danwdart:
Yep awesome that's what I'm looking into atm

eduardoboucas:
Bear in mind this doesn't exist in API yet. It will be released as 4.0 RC1 probably tomorrow, and I'm still writing the documentation as we speak. If you think it's easier to wait until the implementation is fully stable and documentation is available, it's totally fine to put this on hold and tackle something else in the meantime.

danwdart:
I'm continuing on here - what I did is create a client ID and secret field in Users and use that to request documents, optionally waiting for a token to be stored before doing so.

eduardoboucas:
[...]

PR #399, opened 2018-06-26
Hit API directly

danwdart:
Uses API settings to hit the API endpoint directly, rather than through the publish proxy.

Comes close to fixing #396

Still needs:

- 401 detection
- Saving bearer token to localStorage

### DIFF
--- a/frontend/actions/apiActions.js
+++ b/frontend/actions/apiActions.js
@@ -1,6 +1,7 @@
 import * as Constants from 'lib/constants'
 import * as Types from 'actions/actionTypes'
 import apiBridgeClient from 'lib/api-bridge-client'
+import {signOut} from 'actions/userActions'
 
 export function loadApis () {
   return (dispatch, getState) => {
@@ -22,22 +23,36 @@ export function loadApis () {
       })
       .getConfig()
       .then(config => {
-        apiBridgeClient({api}).getCollections().then(({collections}) => {
+        if ('undefined' === typeof config) {
+          return Promise.reject(new Error('Config is undefined'))
+        }
+        apiBridgeClient({api})
+          .getCollections()
+          .then(response => {
+            if ('undefined' === typeof response) {
+              return Promise.reject(new Error('Response is undefined'))
+            }
+            const collections = response.collections
+
+            if ('undefined' === typeof collections) {
+              return Promise.reject(new Error('No collections'))
+            }
+
           // This bundler will be used to get all the collections schemas for
           // this API in bulk.
-          const collectionBundler = apiBridgeClient.getBundler()
+            const collectionBundler = apiBridgeClient.getBundler()
 
-          collections.forEach(collection => {
-            const collectionQuery = apiBridgeClient({
-              api,
-              collection,
-              inBundle: true
-            }).getConfig()
+            collections.forEach(collection => {
+              const collectionQuery = apiBridgeClient({
+                api,
+                collection,
+                inBundle: true
+              }).getConfig()
 
-            collectionBundler.add(collectionQuery)
-          })
+              collectionBundler.add(collectionQuery)
+            })
 
-          collectionBundler.run()
+            collectionBundler.run()
             .then(apiCollections => {
               const isAuthApi = auth.host === api.host && auth.port === api.port
               const mergedCollections = apiCollections
@@ -70,12 +85,19 @@ export function loadApis () {
               }
             })
             .catch(err => {
+              console.log(73, err)
               dispatch(setApiStatus(Constants.STATUS_FAILED))
             })
-        })
+          })
         .catch(err => {
+          console.log(78, err)
           dispatch(setApiStatus(Constants.STATUS_FAILED))
         })
+      })
+      .catch(err => {
+        console.log(83, err)
+        console.log('No config means I should log out.')
+        dispatch(signOut())
       })
     })
   }
--- a/frontend/actions/userActions.js
+++ b/frontend/actions/userActions.js
@@ -4,8 +4,8 @@ import * as Constants from 'lib/constants'
 import * as Types from 'actions/actionTypes'
 import * as documentActions from 'actions/documentActions'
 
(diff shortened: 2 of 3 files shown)


frontend/actions/apiActions.js
```diff
@@ -22,22 +23,36 @@ export function loadApis () {
       })
       .getConfig()
       .then(config => {
-        apiBridgeClient({api}).getCollections().then(({collections}) => {
+        if ('undefined' === typeof config) {
```
mingard: For readability I'd reverse the statement:
```javascript
if (typeof config === 'undefined') {
...
}
```
You could also consider using the `undefined` primitive type, which we've previously used in the app.
```javascript
if (config === undefined) {
...
}
```

frontend/actions/apiActions.js
```diff
@@ -22,22 +23,36 @@ export function loadApis () {
...
+          return Promise.reject(new Error('Config is undefined'))
+        }
+        apiBridgeClient({api})
+          .getCollections()
+          .then(response => {
+            if ('undefined' === typeof response) {
```
mingard: Same notes re `undefined` and readability here.

frontend/actions/apiActions.js
```diff
@@ -22,22 +23,36 @@ export function loadApis () {
...
+              return Promise.reject(new Error('Response is undefined'))
+            }
+            const collections = response.collections
+
+            if ('undefined' === typeof collections) {
```
mingard: Same notes re undefined and readability here.

frontend/actions/apiActions.js
```diff
@@ -70,13 +85,20 @@ export function loadApis () {
               }
             })
             .catch(err => {
+              console.log(73, err)
```
mingard: Remove logs

frontend/actions/apiActions.js
```diff
@@ -70,13 +85,20 @@ export function loadApis () {
...
               dispatch(setApiStatus(Constants.STATUS_FAILED))
             })
-        })
+          })
         .catch(err => {
+          console.log(78, err)
```
mingard: Remove logs

frontend/actions/apiActions.js
```diff
@@ -70,13 +85,20 @@ export function loadApis () {
...
           dispatch(setApiStatus(Constants.STATUS_FAILED))
         })
       })
+      .catch(err => {
+        console.log(83, err)
+        console.log('No config means I should log out.')
```
mingard: Remove logs

frontend/actions/userActions.js
```diff
@@ -180,9 +180,20 @@ export function signIn (email, password) {
...
+      ).catch(
+        err => console.error({err})
+      )
+
+      return user
+    })
+    .then(user => dispatch(setRemoteUser(user)))
+    .catch(response => {
+      console.log(response)
+      console.error(response.error)
```
mingard: Remove logs

frontend/lib/api-bridge-client.js
```diff
@@ -2,8 +2,11 @@
 
 import 'fetch'
 import * as Constants from 'lib/constants'
+import APIWrapper from '@dadi/api-wrapper-core'
+import Cookie from 'js-cookie'
 
-const APIWrapper = require('@dadi/api-wrapper-core')
+// The cookie name to store the bearer token
+const KEY_BEARER_TOKEN = 'bearer_token'
```
mingard: Perhaps I'm missing something, but this seems to be suggesting that the bearer token is to be made available client-side. Isn't that a security vulnerability?

frontend/lib/api-bridge-client.js
```diff
@@ -163,6 +172,35 @@ module.exports.getBundler = () => {
...
+
+export const getNewBearerToken = function getNewBearerToken (user) {
+  const bearerToken = Cookie.get(KEY_BEARER_TOKEN)
+
+  if (bearerToken) {
+    return Promise.resolve(bearerToken)
+  }
+
+  return fetch(
+    'http:\/\/atlas.jolharg.com:8081\/token',
```
mingard: Is this a finished PR? There seems to be a few placeholder variables.

frontend/actions/apiActions.js
danwdart: I'm used to doing static-first, to prevent accidental ='ing when I mean to == or ===.
But if this project requires the opposite I could change things...

frontend/lib/api-bridge-client.js
danwdart: Not finished, it still needed a few changes when I moved on.

frontend/actions/userActions.js
danwdart: Of course, just want it to work and be finished / ready first

frontend/lib/api-bridge-client.js
danwdart: That's what I was told to do - either as a cookie or localstorage - but was asked to make it a cookie - how would it not be, if we're changing to this bearer token method of hitting from client-side?

eduardoboucas: The ACL piece introduced in API 4.0 allows clients (and their bearer tokens) to have a limited set of permissions, so I don't see this as a security vulnerability. Sure, they can intercept the bearer token being used by peeking into the Ajax requests, but it's _their_ token, which will not give them access to anything they can't already do via the Publish interface.

frontend/actions/apiActions.js
mingard: As a rule, we always use `===`, although worth checking with @eduardoboucas incase this has changed. I would definitely reverse this for readability and consistency with the rest of the code :)

frontend/actions/userActions.js
mingard: Okay - I see the in-progress label now 👍

eduardoboucas:
Sorry, closing this in favour of #431! 😢

PR #402, opened 2018-06-28
Make fields' label containers clickable

### LINKED ISSUE #380: Document field select area
### Expected behavior

Select anywhere in the field to focus it and show text cursor for that area indicating this.

### Actual behavior

screen shot 2018-06-06 at 14 36 26

Only this area causes selection

danwdart:
I think this is the "natural" way but please let me know if you think there's a better way.

Feels a bit weird to have a label inside a label - we could make that a span too if you prefer.

Fixes #380


danwdart:
Yeah, will alter the styles - I thought that having anything else might need severe CSS changes - but due to modularity it sounds easy.

danwdart:
How's this?

danwdart:
@eduardoboucas Is that better? Not sure if you have to press yes again or something

PR #403, opened 2018-07-02
contain compact menu in a div with max height

anaumovets:


anaumovets:
This seems to be the easiest way for creating style conditionally.
Changing to `let innerStyle = new Style()` throws error, so I think it's necessary

PR #411, opened 2018-07-10
Fix delete button css changes

danwdart:


frontend/components/Button/Button.css
```diff
@@ -21,7 +21,8 @@
 }
 
 .button-small {
-  font-size: 13px;
+  font-size: 15px;
+  font-weight: 400px;
```
eduardoboucas: Did you mean to use `px` here?

danwdart: Nope, what's that doing there >.>

danwdart:
@eduardoboucas updated

PR #421, opened 2018-07-19
fix: dropdown navs now have a higher z-index than the menu items they came from

### LINKED ISSUE #305: Navigation z-index above menu
When the top level nav extends to a second row, the menu for the first row becomes unusable.
screenshot 2017-08-15 12 24 56

danwdart:
Fixes #305


danwdart:
I think just these are approved are they not?

PR #423, opened 2018-07-25
increase datetimepicker z-index

anaumovets:


anaumovets:
the obscuring footer's z-index is 1000 so I thought 2000 is a good value

Issue #151, opened 2017-04-03
Document counter wrapping occurring even with sufficient horizontal whitespace


mingard:
When looking at a collection with more than one page (showing jump to page input) the current view metrics are wrapping. 

A side note is that the _Go to page_ input is rather large. No reason to allow more than 5 digits. At a pagination limit of 20 that's 1,000,000 records.
<img width="154" alt="screen shot 2017-04-03 at 18 14 27" src="https://cloud.githubusercontent.com/assets/1352469/24622127/c4f08ce4-189b-11e7-97a1-11add6577856.png">

PR #425, opened 2018-07-27
Patch/mobile select reference document button missing

anaumovets:


anaumovets:
The fix has to be careful not to break https://github.com/dadi/publish/issues/151 - that one needs to be tested too. Perhaps interferes with something else I haven't noticed.

PR #434, opened 2018-08-01
Whitelabelling support

danwdart:


app/config-schema.js
```diff
@@ -322,5 +322,23 @@ module.exports = {
         default: true
       }
     }
+  },
+  whitelabel: {
+    availableInFrontend: true,
+    bgImage: {
```
eduardoboucas: Can we change this to `backgroundImage`?

app/config-schema.js
```diff
@@ -322,5 +322,23 @@ module.exports = {
...
+      doc: 'The background image URL',
+      format: String,
+      default: ''
```
eduardoboucas: Perhaps this should default to the URL of the default image we currently use. Then we'd always read this value from `config.get` and we'd either get the path to whatever image should be used.

app/config-schema.js
```diff
@@ -322,5 +322,23 @@ module.exports = {
...
+    },
+    logo: {
+      doc: 'The logo URL',
+      format: String,
+      default: ''
```
eduardoboucas: ^ Same as comment above

app/config-schema.js
```diff
@@ -322,5 +322,23 @@ module.exports = {
         default: true
       }
     }
+  },
+  whitelabel: {
```
eduardoboucas: Are we sure we want to use the term `whitelabel`? /cc @jimlambie

app/config-schema.js
danwdart: Sounds good to me

app/config-schema.js
danwdart: Yep

app/config-schema.js
danwdart: Of course

danwdart:
We're at a "has config, need to change actual code" stage

danwdart:
screenshot_20180808_170917
Something like this then?

danwdart:
Cool, I think most of these broken tests are fixed by another PR, and I think I've still got to check for defaults working but other than that it seems ok functionality wise

danwdart:
I hope you're okay @abovedave !

danwdart:
Yes, most are fixed by https://github.com/dadi/publish/pull/438 - I don't think I've touched the other, must be from another place

danwdart:
The default image requires an overlay but custom images don't. Should I:
a. make an exception in the code that if it detects the default image add the overlay? (I don't like this)
b. make the overlay a config variable with default true but allow false for non-darkened images?
c. alter the actual default image to make it darker and remove overlays completely?

danwdart:
d. remove the overlay but leave the default image so it looks like this:

screenshot_20180809_153432

eduardoboucas:
Good question.

Let's rule out **a**. I'm not a big fan of **b** either, because I feel that adding something so specific to the configuration is unnecessary noise. So I'd say either **d** or we *always* apply an overlay as a design decision.

One for @abovedave or @joverbye to assist with, I'd say.

danwdart:
Looks like always having an overlay makes my first picture look like this:

screenshot_20180809_154408

So, the default overlay that works with the default picture isn't going to work on other pictures universally.

eduardoboucas:
(I meant to say "either **c** (...)" and not "either **d**" (...)" in my comment above 🤦‍♂️)

Agree, so I think the best option is to remove the overlay functionality and leave it up to the developers to choose an image that is appropriate.

danwdart:
That switches properly between them now - so I think that covers the sign in page, which I think was the original scope.

PR #437, opened 2018-08-02
Feat/textbox view options

### LINKED ISSUE #384: Textbox view options
### Expected behavior

* [ ] Box is always full browser height

screen shot 2018-06-06 at 14 48 04
screen shot 2018-06-06 at 14 48 11

* [ ] Fixed number of rows
* [ ] Fit to content

### Actual behavior

Only resizable by dragging bottom right corner

danwdart:
Yes, I know `this.base.style` is TOTALLY the wrong way to do it, but what am I going to do, re-render self? Is that the correct way of doing it in this case?

Fixes #384

### DIFF
--- a/frontend/components/TextInput/TextInput.css
+++ b/frontend/components/TextInput/TextInput.css
@@ -20,7 +20,7 @@
 }
 
 .full-height {
-  height: 100vh;
+  height: calc(100vh - 135px);
 }
 
 .content-height {
--- a/frontend/components/TextInput/TextInput.css
+++ b/frontend/components/TextInput/TextInput.css
@@ -10,3 +10,19 @@
 .input-in-label {
   background-color: inherit;
 }
+
+.resizable {
+  resize: both;
+}
+
+.not-resizable {
+  resize: none;
+}
+
+.full-height {
+  height: 100vh;
+}
+
+.content-height {
+  height: auto;
+}
--- a/frontend/components/TextInput/TextInput.jsx
+++ b/frontend/components/TextInput/TextInput.jsx
@@ -111,7 +111,7 @@ export default class TextInput extends Component {
   }
 
   static defaultProps = {
-    heightType: `static`,
+    heightType: 'static',
     inLabel: false,
     multiline: false,
     readonly: false,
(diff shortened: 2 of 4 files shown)


frontend/components/TextInput/TextInput.jsx
```diff
@@ -104,21 +126,35 @@ export default class TextInput extends Component {
...
     } = this.props
 
     let inputStyle = new Style(styles, 'input')
 
     inputStyle.addIf('input-in-label', inLabel)
+    inputStyle.add(resizable ? `resizable` : `not-resizable`)
+    if (heightType === `full`) {
+      inputStyle.add(`full-height`)
+    }
+    else if (heightType === `content`) {
```
eduardoboucas: `else` in the same line as `if`, please 😇

frontend/components/TextInput/TextInput.jsx
```diff
@@ -104,21 +126,35 @@ export default class TextInput extends Component {
...
+      resizable,
+      rows,
       type,
       value
     } = this.props
 
     let inputStyle = new Style(styles, 'input')
 
     inputStyle.addIf('input-in-label', inLabel)
+    inputStyle.add(resizable ? `resizable` : `not-resizable`)
```
eduardoboucas: Any reason to use template literals in place of regular strings?

danwdart: I'm used to it. No other reason. It's also less work to add templates if I need to, so I've just started to default to it. I'm all for not doing so if that's the rules / we're using `standard` / etc. Let's eslint the lot, really.

frontend/components/TextInput/TextInput.jsx
danwdart: Oh you mean `} else  if (` ? Of course, missed that one.

jimlambie:
http://publish-feat-textbox-view-options.eb.dev.dadi.technology/articles2/new

frontend/components/TextInput/TextInput.css
```diff
@@ -9,4 +9,20 @@
...
+.resizable {
+  resize: both;
+}
+
+.not-resizable {
+  resize: none;
+}
+
+.full-height {
+  height: 100vh;
```
abovedave: Maybe this should account for the save bar e.g., ` calc(100vh - 135px)`?

danwdart: Sounds good, I'll add it

danwdart:
Definitely needs a few more tests (I'm adding)

PR #479, opened 2018-09-05
Attempt to switch autocomplete off

jimlambie:


eduardoboucas:
I'm going to close this one for now, if that's okay. If we're seeing issues with autocomplete, can we please open a ticket with instructions on how to replicate on the preview build or, if not replicable there, a testbed so we can replicate locally?

Thanks! 🍸

PR #484, opened 2018-09-10
Readonly datetime (part 2 - the saga continues)

### LINKED ISSUE #478: DateTime is still editable when set to readonly
### Expected behavior

The field is disabled.

### Actual behavior

You can edit the field as normal

```json
"publishedAt": {
  "type": "DateTime",
  "format": "unix",
  "label": "Publish date/time",
  "publish": {
    "section": "Main",
    "placement": "sidebar",
    "readonly": true,
    "display": {
      "edit": true,
      "list": true
    }
  }
}
```

jimlambie:
Fix #478

### DIFF
--- a/frontend/components/FieldDateTime/FieldDateTimeEdit.jsx
+++ b/frontend/components/FieldDateTime/FieldDateTimeEdit.jsx
@@ -203,8 +203,10 @@ export default class FieldDateTimeEdit extends Component {
     const {pickerVisible} = this.state
     const {schema} = this.props
 
+    const publishBlock = schema.publish || {}
+
     // Return from focus event so picker doesn't display
-    if (schema.publish.readonly === true) {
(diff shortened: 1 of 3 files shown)


frontend/components/FieldDateTime/FieldDateTimeEdit.jsx
```diff
@@ -201,6 +201,12 @@ export default class FieldDateTimeEdit extends Component {
 
   handleFocus(hasFocus) {
     const {pickerVisible} = this.state
+    const {schema} = this.props
+
+    // Return from focus event so picker doesn't display
+    if (schema.publish.readonly === true) {
```
eduardoboucas: Is it worth adding a check for `schema.publish`, in case it doesn't exist? I don't think we are ensuring that property exists anywhere upstream.

PR #486, opened 2018-09-11
Resolve reference fields in list view

### LINKED ISSUE #476: List view: resolve reference fields
### Expected behavior

The same behaviour as in a document i.e., the text from the first `String` field and a link to the referenced document.

(Could we even choose which sub-document field to display?)

### Actual behavior

Currently reference fields are blank if displayed in the list view.

jimlambie:
Close #476

### DIFF
--- a/frontend/components/FieldReference/FieldReferenceList.jsx
+++ b/frontend/components/FieldReference/FieldReferenceList.jsx
@@ -61,19 +61,18 @@ export default class FieldReferenceList extends Component {
 
     const firstStringField = this.findFirstStringField(displayableFields)
 
-    if (optionsBlock) {
-      return (<div class={styles.values}>
-        {value.map(val => {
-          let collection = schema.settings.collection
-          let editLink = `${collection}/${val._id}`
-          let displayField = optionsBlock.field || firstStringField ? firstStringField.key : null
+    return (<div class={styles.values}>
+      {value.map((val, idx) => {
+        console.log(val)
+        let collection = schema.settings.collection
+        let editLink = `${collection}/${val._id}`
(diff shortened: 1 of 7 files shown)


frontend/components/FieldReference/FieldReferenceList.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+    const firstStringField = this.findFirstStringField(displayableFields)
+
+    const values = value && !(value instanceof Array) ? [value] : value
+
+    return (<div class={styles.values}>
+      {values.map((val, idx) => {
+        let collection = schema.settings.collection
+        let editLink = `${collection}/${val._id}`
+        let displayField = optionsBlock && optionsBlock.displayField || firstStringField ? firstStringField.key : null
```
abovedave: Should this live in the `publish.display` block instead?

frontend/components/FieldReference/FieldReferenceList.jsx
```diff
@@ -0,0 +1,81 @@
...
+    const values = value && !(value instanceof Array) ? [value] : value
+
+    if (values) {
+      return (<div class={styles.values}>
+        {values.map((val, idx) => {
+          let collection = schema.settings.collection
+          let editLink = `${collection}/${val._id}`
+          let displayField = optionsBlock && optionsBlock.displayField || firstStringField ? firstStringField.key : null
+
+          return <div><a class={styles['value-link']} href={editLink}>
```
eduardoboucas: We usually return JSX blocks within parenthesis and in a new line, so:

```jsx
return (
  <div>(...)</div>
)
```

Do you mind changing? Same on lines 65 and 73.

PR #495, opened 2018-09-17
Ensure fields can be rendered without optional schema

anaumovets:


anaumovets:
These changes are best tested using storybook, which is not integrated in develop.
To check that fields are working with minimal schema, check out this branch:
https://github.com/dadi/publish/tree/feature/storybook-integration
`npm install` and `npm run-script storybook`

frontend/components/FieldDateTime/FieldDateTimeFilter.jsx
```diff
@@ -76,6 +76,8 @@ export default class FieldDateTimeFilter extends Component {
       valueStyles
     } = this.props
 
+    const format=(((config || {}).formats || {}).date || {}).short
```
eduardoboucas: I understand why you're doing this, but I find this syntax super hard to read and not a justifiable price to pay for a single-liner. Could you refactor?

frontend/components/FieldImage/FieldImageEdit.jsx
```diff
@@ -156,7 +179,7 @@ export default class FieldImageEdit extends Component {
                 <span>or </span>
                 <FileUpload
                   allowDrop={true}
-                  accept={config['FieldImage'].accept}
+                  accept={((config || {})['FieldImage'] || {}).accept}
```
eduardoboucas: Same thing as above.

frontend/components/FieldImage/FieldImageReferenceSelect.jsx
```diff
@@ -72,7 +72,8 @@ export default class FieldImageReferenceSelect extends Component {
   }
 
   render() {
-    const {data} = this.props
+    let {data} = this.props
```
eduardoboucas: Can we simply do:

```
let data = this.props.data || []
```

frontend/components/FieldBoolean/FieldBooleanEdit.jsx
```diff
@@ -83,15 +99,15 @@ export default class FieldBooleanEdit extends Component {
...
 
-    const publishBlock = schema.publish
+    const publishBlock = schema.publish || {}
     const readOnly = publishBlock.readonly === true
 
     return (
       <Label
         compact={true}
-        label={schema.label}
+        label={displayName || ''}
```
eduardoboucas: I see this pattern (`displayName || ''`) repeated across this PR. I believe that is because of [this check](https://github.com/dadi/publish/blob/develop/frontend/components/Label/Label.jsx#L121) in the `Label` component, that stops the component (and its children) from being rendered if the `label` param is falsy.

Instead of supplying a fallback each time we call `Label`, should we add the fallback to the component instead? I'm not sure what the behaviour would be like, since we're talking about rendering a label without any text. Perhaps the logic should be to simply render the children without a wrapping `<label>` element when the `label` prop isn't a String?

Either way, I would feel more comfortable in handling that case in the component itself, rather than ensuring we supply a fallback each time. What do you think?

frontend/components/FieldImage/FieldImageReferenceSelect.jsx
```diff
@@ -93,8 +94,6 @@ export default class FieldImageReferenceSelect extends Component {
         ))}
       </div>
     )
-
-    return this.renderColumns(data)
```
eduardoboucas: Nice catch! 😬

anaumovets:
@eduardoboucas , everything fixed according to your comments.
As for the label, I decided to set label to empty string and render as previously (but move this inside the Label component as you suggested). Rendering children without wrapping could affect their display styling, so I decided to still have them wrapped.

frontend/components/FieldDateTime/FieldDateTimeFilter.jsx
```diff
@@ -67,7 +67,7 @@ export default class FieldDateTimeFilter extends Component {
   }
 
   render() {
-    const {
+    let {
       analyserStyles,
       config,
```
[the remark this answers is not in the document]
anaumovets: smart stuff

Issue #435, opened 2018-08-02
Ensure fields are rendered if optional schema properties are missing


lshchur:
Re: https://github.com/dadi/publish/blob/6ffe829feefee5a0202ae5926fa502c75fa9b92e/frontend/components/FieldString/FieldStringEdit.jsx#L127

James: we need a ticket for defaulting to the name of the field when label is missing

danwdart:
Is this for all FieldString/FieldNumber etc?
I'd suggest that we further abstract a generic FieldText and have FieldString/FieldNumber just be validators for it maybe - different from TextBox - I'm having to do something very similar with it here and don't want to step on everybody's toes

lshchur:
It's actual for FiledNumber, yes, because in it, it's also wrapped by Label which is given a prop `label={schema.label}`

With FieldImage, for example, it's a bit different. In there the passed label equals ```schema.label || schema._id```, so there's some kind of a fallback value in case if `label` is not specified
With FieldPassword it's also different, just for the record
```label={`Current ${schema.label.toLowerCase()}`}```

danwdart:
Sorry my comment went on the wrong issue somehow

danwdart:
So should `schema.label || schema._id` be passed to the rest?

lshchur:
Yes, that sounds about right.

eduardoboucas:
Just wanted to add that Publish should be able to render any field as long as it has a `type` property. Everything else should be optional.

If that's okay, I'm going to change the title of this ticket so the scope becomes ensuring the above is respected.

PR #496, opened 2018-09-17
Fix flashing upload drop zone

### LINKED ISSUE #416: File upload drop target issue
### Expected behavior
Should be a way to drop files into a file upload

### Actual behavior
The drop target changes quickly and makes it difficult to drop

Drop a file onto the drop target on an upload screen!

### Package details
`develop` branch once #393 is closed / #415 is merged

danwdart:
Fixes #416


danwdart:
There are many unrelated tests breaking

eduardoboucas:
As discussed on the Publish call, I'm happy to merge this and we can revisit if we're still not happy with the result.

I see some merge conflicts now, is this something you could fix?

danwdart:
Yep of course

eduardoboucas:
I think the change was done by @anaumovets as part of #435. From what I gather, the newly-created `accept` variable is the same as drilling into `config`, except it's safeguarded against missing schema properties.

If that doesn't make sense, it's probably worth asking him.

danwdart:
Okay that did it

PR #502, opened 2018-09-19
Save the Save Options

### LINKED ISSUE #379: Document save options drop-down: remember previously selected
screen shot 2018-06-06 at 14 33 14

### Expected behavior

Per-user remember the previous state of this button - handy when bulk entering data.

### Actual behavior

Always reverts to "save and continue"

danwdart:
Fixes #379

### DIFF
--- a/docs/components/ButtonWithOptions.md
+++ b/docs/components/ButtonWithOptions.md
@@ -56,7 +56,7 @@ Object containing the secondary options. Keys define the text of the option and 
  ```jsx
  <ButtonWithOptions
    onClick={this.mainCallback()}
-   options={new Map(
+   options={[
      [
       'Save and continue',
        {
@@ -67,13 +67,13 @@ Object containing the secondary options. Keys define the text of the option and 
          ]
        }
      ]
-   ])}
+   ]}
  >
    Save
  </ButtonWithOptions>
  ```
 
-- type: `custom`
+- type: `array`
 
 
 ### `type`
--- a/docs/components/ButtonWithOptions.md
+++ b/docs/components/ButtonWithOptions.md
@@ -13,6 +13,13 @@ Colour accent.
 - type: `enum('system')`
 
 
+### `callback`
+
+function to call with the value of the action
+
+- type: `func`
+
+
 ### `children`
 
 The text to be rendered inside the main button.
@@ -49,16 +56,24 @@ Object containing the secondary options. Keys define the text of the option and 
  ```jsx
  <ButtonWithOptions
    onClick={this.mainCallback()}
-   options={{
-     'Save and continue': this.saveAndContinueCallback()
-     'Save and go back': this.saveAndGoBackeCallback()
-   }}
+   options={new Map(
+     [
+      'Save and continue',
+       {
+         action: 'save',
+         methods: [
+           'edit',
+           'new'
+         ]
+       }
+     ]
+   ])}
  >
    Save
  </ButtonWithOptions>
  ```
 
-- type: `object`
+- type: `custom`
 
 
 ### `type`
--- a/frontend/components/ButtonWithOptions/ButtonWithOptions.jsx
+++ b/frontend/components/ButtonWithOptions/ButtonWithOptions.jsx
@@ -48,7 +48,7 @@ export default class ButtonWithOptions extends Component {
      *  ```jsx
      *  <ButtonWithOptions
      *    onClick={this.mainCallback()}
-     *    options={new Map(
+     *    options={[
      *      [
      *       'Save and continue',
      *        {
@@ -59,13 +59,13 @@ export default class ButtonWithOptions extends Component {
      *          ]
      *        }
      *      ]
-     *    ])}
+     *    ]}
      *  >
      *    Save
      *  </ButtonWithOptions>
      *  ```
      */
-    options: proptypes.map,
+    options: proptypes.array,
 
     /**
      * function to call with the value of the action
@@ -104,12 +104,12 @@ export default class ButtonWithOptions extends Component {
   render() {
     const {
       accent,
+      callback,
       children,
       disabled,
       href,
       onClick,
       options,
-      callback,
       type
     } = this.props
     const {open} = this.state
@@ -148,8 +148,8 @@ export default class ButtonWithOptions extends Component {
         {open && !disabled &&
           <div class={styles.options}>
             <Dropdown tooltip="right">
-              {Array.from(options.keys()).map(name =>
-                  <DropdownItem onClick={event => callback(name)}>{name}</DropdownItem>
+              {options.map(([name]) =>
+                  <DropdownItem onClick={() => callback(name)}>{name}</DropdownItem>
               )}
             </Dropdown>
           </div>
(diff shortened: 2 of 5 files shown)


docs/components/ButtonWithOptions.md
```diff
@@ -49,16 +56,24 @@ Object containing the secondary options. Keys define the text of the option and
  ```jsx
  <ButtonWithOptions
    onClick={this.mainCallback()}
-   options={{
-     'Save and continue': this.saveAndContinueCallback()
-     'Save and go back': this.saveAndGoBackeCallback()
-   }}
+   options={new Map(
```
eduardoboucas: I understand that a Map is semantically more correct than just a plain object, but I'm not keen on the complexity it adds (lengthier declaration + `Array.from(map.keys())`). The reason we went with an object in the first place was for simplicity of directly mapping keys (labels) to values (callbacks). If you think an object is ugly in this case, I'd rather go with an array containing objects with `label` and `callback` properties.

What do you think? Was there any other reason for this change that I might be overlooking?

frontend/components/ButtonWithOptions/ButtonWithOptions.jsx
```diff
@@ -96,6 +109,7 @@ export default class ButtonWithOptions extends Component {
       href,
       onClick,
       options,
+      callback,
```
eduardoboucas: Can you maintain alphabetical order of the props, please? 😇

frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
```diff
@@ -24,6 +25,71 @@ import DropdownNative from 'components/DropdownNative/DropdownNative'
 import Peer from 'components/Peer/Peer'
 import Toolbar from 'components/Toolbar/Toolbar'
 
+const FIELD_PUBLISH_SAVE_OPTIONS_LAST_USED = 'publishSaveOptionsLastUsed',
```
eduardoboucas: I know it's less bloated, but can we avoid comma-separated declarations and have one `const`/`let` per variable?

frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
```diff
@@ -24,6 +25,71 @@ import DropdownNative from 'components/DropdownNative/DropdownNative'
...
+  SAVE_AND_CONTINUE = 'Save and continue',
```
eduardoboucas: I understand why we have the mode identifiers (e.g. `save`, `saveAndCreateNew`) defined as constants, since we'll use them on the interface and also save them to API. But do we need constants for the labels and for `publishSaveOptionsLastUsed`?

frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
```diff
@@ -137,17 +203,23 @@ class DocumentEditToolbar extends Component {
...
+    const savedSaveOptionOnUser = state.user.remote.data[FIELD_PUBLISH_SAVE_OPTIONS_LAST_USED]
 
-    // If we're editing an existing document, we also allow users to duplicate
-    // the document.
-    if (method === 'edit') {
-      saveOptions['Save as duplicate'] = this.handleSave.bind(this, 'saveAsDuplicate')
-    }
+    const savedSaveOption = savedSaveOptionOnUser || availableSaveOptions.default
+
+    const localSaveOptions = filterMap(
```
eduardoboucas: Again, not a huge fan of the usage of `Map` here. All we're doing is finding a match in an array-like structure, and yet we end up with this somewhat cryptic expression. Do you think we could use an array and `Array.prototype.find`?

frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
```diff
@@ -228,10 +300,11 @@ class DocumentEditToolbar extends Component {
             <ButtonWithOptions
               accent="save"
               disabled={hasConnectionIssues || hasValidationErrors || isSaving}
-              onClick={this.handleSave.bind(this, 'save')}
-              options={saveOptions}
+              onClick={this.saveDefaultActionAndExecute.bind(this, defaultSaveOptionName)}
+              options={invisibleSaveOptions}
+              callback={this.saveDefaultActionAndExecute.bind(this)}
```
eduardoboucas: Maintain alphabetical order of props, please.

frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
```diff
@@ -375,6 +448,38 @@ class DocumentEditToolbar extends Component {
...
+   * @param {string} defaultActionName
+   * @return void
+   */
+  saveDefaultAction(defaultActionName) {
+    const {
+      actions,
+      state
+    } = this.props
+
+    actions.updateLocalUser(`data.${FIELD_PUBLISH_SAVE_OPTIONS_LAST_USED}`, defaultActionName);
```
eduardoboucas: Did I spot a... semi-colon? 😱

🚨 🚨 🚨

frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
danwdart: well, moving to using `Array.prototype.filter` as it was anyway, for filtering stuff with

frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
danwdart: What would our rules be for constants then? Should I continue separation or only use constants when two or more things use them?

docs/components/ButtonWithOptions.md
danwdart: Changed to use the array form of the map I was using - hope that's ok

frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
danwdart: Yep done

frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
danwdart: That's been pushed

frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
```diff
@@ -249,7 +315,7 @@ class DocumentEditToolbar extends Component {
       collection,
       group,
       referencedField,
-      state
+      state,
```
eduardoboucas: 🤔

frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
```diff
@@ -282,7 +348,7 @@ class DocumentEditToolbar extends Component {
       group,
       onBuildBaseUrl,
       section,
-      state
+      state,
```
eduardoboucas: 🤔

frontend/containers/DocumentEditToolbar/DocumentEditToolbar.jsx
danwdart: that's a weird one that crept in refactors, needs a good eslint

danwdart:
That should be better

Issue #385, opened 2018-06-06
Media collections


abovedave:
### Expected behavior

Be able to view media collection and edit/delete/add files (image to start?)

![screen shot 2018-06-06 at 14 57 49](https://user-images.githubusercontent.com/649231/41042613-0518b7e2-699a-11e8-8fe7-75f71b3a84ee.png)

### Actual behavior

You cannot see any media collections

danwdart:
The cloud images collection did look a little like this with some CSS tweaking. I suppose one needs to enable media in the menu first though.

danwdart:
Having a look to see if I can enable this atm (i.e. this afternoon)

danwdart:
cc @abovedave @eduardoboucas

danwdart:
Had a lot of trouble with this - but this is what I considered:
* Publish show menu item for media
* Publish route for media collection(s) should already exist but doesn't go anywhere - is it a bug?
* Pretend that the API gave it in the list of collections - but seems hacky

eduardoboucas:
> Pretend that the API gave it in the list of collections - but seems hacky

Not sure if this helps, but Publish uses the `/api/collections` endpoint from API to populate the list of collections, and this endpoint makes a clear distinction of normal vs. media collections. Example:

```json
{
    "collections": [
        {
            "version": "vjoin",
            "database": "testdb",
            "name": "books",
            "slug": "books",
            "path": "/vjoin/testdb/books"
        },
        {
            "version": "vjoin",
            "database": "testdb",
            "name": "users",
            "slug": "users",
            "path": "/vjoin/testdb/users"
        },
        {
            "version": "vtest",
            "database": "testdb",
            "name": "test-schema",
            "slug": "test-schema",
            "path": "/vtest/testdb/test-schema"
        }
    ],
    "media": {
        "buckets": [
            "mediaStore"
        ],
        "defaultBucket": "mediaStore"
    }
}
```

> Publish show menu item for media

I think it's sensible for Publish to display the media collections in the nav, but not sure where/how. Does it add a new top-level menu called "Media" at the end of the list of collections?

> Publish route for media collection(s) should already exist but doesn't go anywhere - is it a bug?

Where do you see that? API has changed the way it handles media recently. We used to allow normal collections to become media collections by using a `type` property. That's no longer the case. Maybe what you're seeing is an old collection.

danwdart:
[...]

PR #518, opened 2018-10-01
Add media list view

eduardoboucas:
@danwdart has taken this over and has edited this comment.
Fixes https://github.com/dadi/publish/issues/385


danwdart:
@eduardoboucas Things not sure about:
1. Filters - should I remove until they're made properly?
1. Upload/delete many images needs API - atm just using promise all
1. Actions still use fetch - they could use things in api-wrapper instead
1. Bucket is hardcoded to mediaStore atm (route should probably pass in) - when we want to start supporting multiple buckets properly I can fix it (not sure whether as yet)
1. Don't know what to do with clicking images in the grid list view. Show in modal? Show in separate view? Allow metadata editing in such a view? Haven't had a design for that, so if we want it to do something, let's request one

I think that's it

frontend/actions/documentsActions.js
```diff
@@ -33,6 +33,41 @@ export function deleteDocuments ({api, collection, ids}) {
...
+              'Authorization': `Bearer ${getState().user.accessToken}`
+            },
+            method: 'DELETE'
+          }
+        )
+      )
+    )
+    .then(
+      responses => {
+        dispatch(
```
eduardoboucas: You're dispatching the "is deleting" action after `Promise.all` resolves, at which point all items have been deleted?

frontend/components/FileUploadMedia/FileUploadMedia.jsx
```diff
@@ -0,0 +1,90 @@
...
+        <div class={styles['upload-options']}>
+          <DropArea
+            draggingText={`Drop image(s) here`}
+            onDrop={this.handleFileChange.bind(this)}
+          >
+            <div class={styles['upload-drop']}>
+              Drop file(s) to upload
+            </div>
+          </DropArea>
+        </div>
```
eduardoboucas:
```suggestion
        </div>

```

frontend/components/FileUploadMedia/FileUploadMedia.jsx
```diff
@@ -0,0 +1,90 @@
...
+        <div class={styles['upload-select']}>
+          <span>or </span>
```
eduardoboucas:
```suggestion
          <span>or </span>

```

frontend/components/FileUploadMedia/FileUploadMedia.jsx
```diff
@@ -0,0 +1,90 @@
...
+          <FileUpload
```
eduardoboucas: Can you sort props alphabetically, please?

frontend/components/FileUploadMedia/FileUploadMedia.jsx
```diff
@@ -0,0 +1,90 @@
...
+  handleFileChange(files) {
+    const {
+      actions: {
+        uploadMediaToBucket
+      },
+      bucket,
+      state: {
+        api: {
+          apis: [
+            api
```
eduardoboucas: Hmm, I'm not sure I'm a fan of drilling down this deep into the object using the destructuring syntax. For consistency with the rest of the codebase, I'd stick to destructuring one level at most and then repeating the process using other variables if needed.

frontend/actions/documentActions.js
```diff
@@ -124,6 +125,56 @@ export function registerUserLeavingDocument ({
...
+  bucket,
+  files
+}) {
+  return (dispatch, getState) => {
+    dispatch(
+      setRemoteDocumentStatus(Constants.STATUS_SAVING)
+    )
+
+    Promise.all(
+      Array.from(files).map(
```
eduardoboucas: Could you make the function argument part of the same line as the function name to avoid an unnecessary level of indentation?

frontend/actions/documentActions.js
```diff
@@ -124,6 +125,56 @@ export function registerUserLeavingDocument ({
...
+                'Authorization': `Bearer ${getState().user.accessToken}`
+              },
+              method: 'POST'
+            }
+          )
+        }
+      )
+    )
+    .then(
+      () =>
```
eduardoboucas: Could you make the function argument part of the same line as the function name to avoid an unnecessary level of indentation?

frontend/actions/documentsActions.js
```diff
@@ -33,6 +33,41 @@ export function deleteDocuments ({api, collection, ids}) {
...
+              'Accept': 'application/json',
+              'Authorization': `Bearer ${getState().user.accessToken}`
+            },
+            method: 'DELETE'
+          }
+        )
+      )
+    )
+    .then(
+      responses => {
```
eduardoboucas: Could you make the function argument part of the same line as the function name to avoid an unnecessary level of indentation?

frontend/components/FileUploadMedia/FileUploadMedia.jsx
```diff
@@ -0,0 +1,90 @@
...
+          ]
+        }
+      }
+    } = this.props
+
+    uploadMediaToBucket(
+      {
```
eduardoboucas: Do you mind making the opening curly braces part of the same line to avoid an unnecessary level of indentation?

```js
uploadMediaToBucket({
  api,
  bucket,
  files
})
```

frontend/containers/DocumentListToolbar/DocumentListToolbar.jsx
```diff
@@ -215,11 +215,19 @@ class DocumentListToolbar extends Component {
     } = this.props
 
     if (bulkActionSelected === 'delete') {
-      actions.deleteDocuments({
-        api,
-        collection,
-        ids: state.documents.selected
-      })
+      if (collection == "mediaStore") {
```
eduardoboucas: Can we use a strong comparison ( `===`)? Also, is `collection` not an object that may or may not have a `_media` property to identify it as a media collection? I'd rather not hardcode the bucket name here.

frontend/index.jsx
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
 import {h, render} from 'preact'
+import 'preact/debug'
```
eduardoboucas: What is this one about?

frontend/views/MediaListView/MediaListView.jsx
```diff
@@ -0,0 +1,112 @@
...
+        fileName: {
+          label: 'Filename'
+        }
+      },
+      name: 'Media'
+    }
+
+    return (
+      <Page>
+        <Header/>
```
eduardoboucas:
```suggestion
        <Header/>

```

frontend/components/FileUploadMedia/FileUploadMedia.jsx
danwdart: Alright, I'll use `const api = state.api.apis[0]

frontend/actions/documentsActions.js
danwdart: Ah, so I should use that first, and then when it resolves use a "has deleted", right?

frontend/components/FileUploadMedia/FileUploadMedia.jsx
danwdart: Huh? What does this mean? Do you mean we should have empty lines at the end of each file?

frontend/actions/documentsActions.js
danwdart: This in general I shall try to stick to.

frontend/components/FileUploadMedia/FileUploadMedia.jsx
danwdart: I should get into that practice.

frontend/components/FileUploadMedia/FileUploadMedia.jsx
eduardoboucas: This is GitHub being weird. I was just suggesting a blank line between components.

frontend/containers/DocumentListToolbar/DocumentListToolbar.jsx
danwdart: Oops - missed that - I always do strong comparisons as a matter of habit and must have missed that.
Yep, just going to pass in the name of the bucket from above - in the above component we've passed a `currentCollection` string it turns out.

frontend/index.jsx
danwdart: Didn't mean to commit that sorry

danwdart: Just enables the react devtools extension for local

danwdart:
About that package-lock, are we not meant to keep that as a matter of making sure we don't have any accidental breaking changes? Would you rather just use the specified versions exactly and gitignore it?

frontend/actions/documentsActions.js
eduardoboucas: See the [deleteDocuments action](https://github.com/dadi/publish/blob/2061a86bd28d1452a4d7d7c12f5052104c99c2ca/frontend/actions/documentsActions.js#L15-L34) example. We set the state to "saving" before we start the operation. The action that is fired after the operation is complete will make the reducer set `isDeleting` back to false, which the components can listen to to know when a delete operation has finished (e.g. `previousState.isDeleting && !currentState.isDeleting`).

frontend/containers/DocumentListToolbar/DocumentListToolbar.jsx
eduardoboucas: Ah, yes. That changed as part of my refactoring. Do you mind updating the `propTypes` to reflect that?

frontend/index.jsx
eduardoboucas: Ah, cool!

danwdart: Well, it enables it globally here, so we should maybe only enable it for local if wanted, or keep it if wanted. I just generally keep it uncommitted in my tree like this.

danwdart: After all it only works when NODE_ENV=development anyway so I don't really see the harm

eduardoboucas:
> About that package-lock, are we not meant to keep that as a matter of making sure we don't have any accidental breaking changes? Would you rather just use the specified versions exactly and gitignore it?

In theory, yes, but we had some problems with it in the past so we're avoid committing for now. We can revisit later though.

eduardoboucas:
The text styling here looks a bit off?

screen shot 2018-10-18 at 14 11 09

frontend/containers/DocumentListToolbar/DocumentListToolbar.jsx
danwdart: Actually it was an object after all in this context, it was made into one from a string.
Anyway I'll introduce an "isMedia" to avoid scanning for media collections and just let the view pass in true.

frontend/index.jsx
danwdart: Would you be okay with keeping it in that case?

eduardoboucas: Yep, it sounds fine.

danwdart:
Currently fixing that styling

danwdart:
Font fixed now - was there anything else we needed to touch in this tier?

eduardoboucas:
I can't get uploads to work:

```
Uncaught ReferenceError: setRemoteDocumentStatus is not defined
    at bundle.js:1
    at Object.dispatch (bundle.js:1)
    at dispatch (<anonymous>:1:28684)
    at bundle.js:1
    at FileUploadMedia.value (bundle.js:5)
    at FileUpload.value (bundle.js:1)
    at HTMLDivElement.eventProxy (bundle.js:1)
bundle.js:1 Uncaught TypeError: component._renderCallbacks.pop(...).call is not a function
    at renderComponent (bundle.js:1)
    at GridList.forceUpdate (bundle.js:1)
    at bundle.js:1
```

Plus the flickering of the drop zone is back.

^ @danwdart

Issue #453, opened 2018-08-21
Add loading spinner after sign in


eduardoboucas:
The app appears unresponsive for a second or so whilst the user is being authenticated. Some visual feedback indicating that the app is performing an action would be useful.

jimlambie:
Closed by https://github.com/dadi/publish/pull/523

PR #523, opened 2018-10-04
Feature/ui tweaks

abovedave:
* Language dropdown hover style (also arrow direction matches other similar elements)
* New list view storting styles
* Loading overlay below nav so it's use is not blocked if something is slow
* Blue to match the newer dadi.cloud blue 
* Tiny fix to truncated strings in list view if broken with a space before the `…`


eduardoboucas:
Looks like a million bucks! 🏆

jimlambie:
@abovedave & @eduardoboucas does this resolve the "loading spinner" issue #453 ?

PR #551, opened 2018-10-19
Improve router's use of URL params

### LINKED ISSUE #545: Pagination links incorrect when sorting
### Expected behavior

A pagination link for a sorted collection should be like:
`"/articles/2?sort=published&order=asc"`

Clicking a pagination link should load the next set of results

### Actual behavior

`"/articles/2?sort=published&order=asc?sort=published&order=asc"`

When clicking a pagination link, a white screen appears. The following is in the console:

screen shot 2018-10-15 at 17 34 49

Sort the articles document list by any field, then click a pagination link

### Package details
1.0.12

jimlambie:
This PR removes incorrect duplication of querystring parameters from the pagination links when sorting documents in the list view.

Note that the url-helper file has become URLParams with methods `toObject()` and `toString()` to better match everything else in the universe.

It also modifies the health check initialisation so that it only occurs once.

Close #545 

**NOTE**
This PR also removes the `TZ` and `assets` properties from the config schema, and adds a validation check on the `server.healthcheck.frequency` setting (rather than checking in code).


frontend/lib/util/urlParams.js
```diff
@@ -0,0 +1,53 @@
...
+    this.input = input
+  }
+
+  toObject () {
+    if (!this.input || typeof this.input !== 'string') {
+      return
+    }
+
+    let params = this.input.replace(/^(\?)/, '')
+              .replace(/"/g, '\\"')
```
eduardoboucas:
```suggestion
      .replace(/"/g, '\\"')
```

frontend/lib/util/urlParams.js
```diff
@@ -0,0 +1,53 @@
...
+              .replace(/&/g, '","')
```
eduardoboucas:
```suggestion
      .replace(/&/g, '","')
```

frontend/lib/util/urlParams.js
```diff
@@ -0,0 +1,53 @@
...
+              .replace(/=/g, '":"')
```
eduardoboucas:
```suggestion
      .replace(/=/g, '":"')
```

PR #552, opened 2018-10-19
image fields must only accept image files for upload

### LINKED ISSUE #543: Ensure DropArea respects MIME type validation
If the user tries to upload a non-image file, the field should become red (similar to invalid input in text fields).

anaumovets:
Fix #543


anaumovets:
There's still one little problem with this: when the images in FileImageEdit field are changed via reference image selection, the error is not removed until you refresh. Thus if you try to upload a nonimage and then select a correct reference image, error will not disappear. I didn't find a solution that would look good to me, that is without coupling these components too much.
Also, I merged latest develop, but there were still some errors in the router when running `npm t` (I didn't touch the router).

eduardoboucas:
This has been incorporated in #554.

PR #538, opened 2018-10-12
Render first image and quantity info for multiple image reference fields

anaumovets:
Close #536

PR #554, opened 2018-10-23
Media upload field

jimlambie:
This PR introduces the idea of a Media field designed to replace the Image field so that it can be used for multiple file types. Accepted file types (default is all files `["*/*"]`) are configured using the API collection schema, for example:

```
"validation": {
  "mimeTypes": ["application/pdf", "image/*"]
}
```

This is under construction, still needs to have error message logic from @anaumovets rolled into it

**Testing**
Use the "Team" collection: http://publish-feature-media-upload-field.eb.dev.dadi.technology/content/team/a3de7ead-80ed-46b3-8008-ad1b5b9ce91b

**Note**
Incorporates PRs #526 and #538 

Close #526 
Close #538


frontend/components/FieldMedia/FieldMediaItem.jsx
```diff
@@ -0,0 +1,68 @@
...
+    if (value.mimetype && value.mimetype.indexOf('image') > -1) {
+      let src = ''
+
+      if (value._previewData) {
+        src = value._previewData
+      } else if (value.url) {
+        src = value.url
+      } else if (value.path) {
+        if (cdn && cdn.publicUrl) {
+          src = `${cdn.publicUrl}/${value.path}`
```
abovedave: Do you think we should add `?width=80` here @jimlambie?

jimlambie: I don't see why not!

frontend/components/FieldMedia/FieldMediaEdit.jsx
```diff
@@ -137,103 +137,131 @@ export default class FieldImageEdit extends Component {
...
-                  src={this.getImageSrc(value)}
+          values.map((value, idx) => {
+
+            let file = {}
+            file.name = value.fileName.split('.').slice(0, -1).join('.')
+            file.size = fileSize(value.contentLength).human('si') || ''
+            file.ext  = value.fileName.split('.').pop()
+
+            let styleValueContainer = new Style(styles, 'value-container')
+              .addIf('value-container-first', values.length > 1 && idx === 0)
```
eduardoboucas: Instead of styling based on classes that are added if an element is the first, the last or the only one, could we not do it using CSS only using `:first-child`, `:last-child` and `:first-child:last-child`?

frontend/components/FieldMedia/FieldMediaEdit.jsx
```diff
@@ -137,103 +137,131 @@ export default class FieldImageEdit extends Component {
...
         {values &&
-          <div class={styles['value-container']}>
-            <div class={styles.thumbnails}>
-              {values.map(value => (
-                <img
-                  class={styles.thumbnail}
-                  src={this.getImageSrc(value)}
+          values.map((value, idx) => {
+
+            let file = {}
```
eduardoboucas: Just curious, could we not declare all three properties as part of the variable initialisation?

frontend/components/FieldMedia/FieldMediaEdit.jsx
```diff
@@ -137,103 +137,131 @@ export default class FieldImageEdit extends Component {
...
+      return this.handleFileChange([files[0]])
     }
-  }
-  
-  handleRemoveFile() {
-    const {name, onChange, schema} = this.props
 
-    if (typeof onChange === 'function') {
-      onChange.call(this, name, null)
+    let values = []
```
eduardoboucas:
```suggestion
    let values = []

```

frontend/components/FieldMedia/FieldMediaEdit.jsx
```diff
@@ -137,103 +137,131 @@ export default class FieldImageEdit extends Component {
...
+    if (value) {
+      values = Array.isArray(value) ? value : [value]
     }
+
+    // Filter for uniqueness by file name and concat.
+    const fileNames = values.map(value => value.fileName)
```
eduardoboucas:
```suggestion
    const fileNames = values.map(value => value.fileName)

```

frontend/components/FieldMedia/FieldMediaEdit.jsx
```diff
@@ -137,103 +137,131 @@ export default class FieldImageEdit extends Component {
...
+    processedFiles = processedFiles.filter(value => !fileNames.includes(value.fileName))
+    onChange.call(this, name,  values.concat(processedFiles))
```
eduardoboucas:
```suggestion
    onChange.call(this, name, values.concat(processedFiles))
```

frontend/components/FieldMedia/FieldMediaEdit.jsx
```diff
@@ -137,103 +137,131 @@ export default class FieldImageEdit extends Component {
...
+    }
 
-    if (value._previewData) return value._previewData
+    if (typeof onChange === 'function') {
+      onChange.call(this, name, newValues)
+    }
+  }
 
-    if (value.url) return value.url
+  handleAddFiles(files) {
```
eduardoboucas: We don't seem to be using this method anywhere and most of its logic seems to be repeated on line 288.

frontend/components/FieldMedia/FieldMediaEdit.jsx
```diff
@@ -252,13 +280,19 @@ export default class FieldImageEdit extends Component {
...
         ) {
-          const finalFiles = singleFile ? processedFiles[0] : processedFiles
 
-          onChange.call(this, name, finalFiles)
+          if (singleFile) {
+            return onChange.call(this, name, processedFiles[0])
+          }
+
+          // Filter for uniqueness by file name and concat.
+          const fileNames = values.map(value => value.fileName)
```
eduardoboucas:
```suggestion
          const fileNames = values.map(value => value.fileName)

```

frontend/components/FieldMedia/FieldMediaItem.jsx
```diff
@@ -0,0 +1,61 @@
...
+    const cdn = config ? config.cdn : null
+
+    if (!value) return null
+
+    const styleFile = new Style(styles, 'file')
+      .addIf('file-list', isList)
+
+    // Get the image path if applicable
+    if (value.mimetype && value.mimetype.indexOf('image') > -1) {
+
```
eduardoboucas:
```suggestion
```

frontend/components/FieldMedia/FieldMediaItem.jsx
```diff
@@ -0,0 +1,61 @@
...
+      if (value.path && cdn && cdn.publicUrl) {
+        src = `${cdn.publicUrl}/${value.path}?width=80`
+      }
+
+      return (
+        <div class={styles.image}>
+          <img src={src} />
+        </div>
+      )
+    } else {
```
eduardoboucas: This `else` is redundant, I believe, as you can just move the `return` to the top level?

frontend/components/FieldMedia/FieldMediaItem.jsx
```diff
@@ -0,0 +1,61 @@
...
+     */
+    value: proptypes.bool
+  }
+
+  constructor(props) {
+    super(props)
+  }
+
+  getSource () {
+    const {config, isList, value} = this.props
```
eduardoboucas: Can you add the `isList` prop to `propTypes`?

frontend/components/FieldMedia/FieldMediaEdit.jsx
```diff
@@ -124,8 +125,7 @@ export default class FieldImageEdit extends Component {
       value
     } = this.props
 
-    const fieldImage = config.FieldImage || {}
-    const accept = fieldImage.accept
+    const accept = schema.validation && schema.validation.accept || ['*/*']
```
eduardoboucas: Is `accept` the right name for this validation operator? Wouldn't `validation.mimeType` be more descriptive?

frontend/components/FieldMedia/FieldMediaEdit.jsx
abovedave: I tried those (also in our case of the mocku pagination button), but `:first-child` seem to have no effect. I think they are only for targeting given element _not_ elements by class name.

jimlambie: 👍

eduardoboucas: That's correct. `:first-child` will target the first element amongst its siblings, as shown by the screenshot:

screenshot 2018-10-24 at 15 57 12

The problem here is that the items we're looking to target aren't the only siblings under the same parent, so `:last-child` will target the wrong element.

You could solve it by creating a `div` to group only the relevant items and that way make use of `:first-child` and `:last-child`. I would actually prefer it this way, but I'm not fussed. Happy for it to be left as is!

PR #559, opened 2018-10-25
first string field takes preference as document link

abovedave:
This PR addresses a small issue if you set a `Reference` field or `Media` field as the first displayed field in your API schema you are unable to click through from the list view (both those fields override the link).

It finds the first field of type `String` and appends the document link to that value, even if it's not in position `0`. If there are no `String` fields in the schema, it fallsback to position `0`

### DIFF
--- a/frontend/containers/DocumentList/DocumentList.jsx
+++ b/frontend/containers/DocumentList/DocumentList.jsx
@@ -295,7 +295,14 @@ class DocumentList extends Component {
     let fieldSchema = collection.fields[column.id]
     let renderedValue = this.renderField(column.id, fieldSchema, value)
 
-    if (index === 0) {
+    let firstStringField = Object.keys(collection.fields).filter(i => {
+      return collection.fields[i].type === 'String'
+    })[0]
+
+    if (
+      (firstStringField && firstStringField === column.id) || 
+      (!firstStringField && index === 0)
+    ) {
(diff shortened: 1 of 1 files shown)


frontend/containers/DocumentList/DocumentList.jsx
```diff
@@ -295,7 +295,14 @@ class DocumentList extends Component {
     let fieldSchema = collection.fields[column.id]
     let renderedValue = this.renderField(column.id, fieldSchema, value)
 
-    if (index === 0) {
+    let firstStringField = Object.keys(collection.fields).filter(i => {
```
eduardoboucas:
```suggestion
    let firstStringField = Object.keys(collection.fields).filter(field => {
```

frontend/containers/DocumentList/DocumentList.jsx
```diff
@@ -295,7 +295,14 @@ class DocumentList extends Component {
...
+      return collection.fields[i].type === 'String'
```
eduardoboucas:
```suggestion
      return collection.fields[field].type === 'String'
```

frontend/containers/DocumentList/DocumentList.jsx
abovedave: Very pedantic, I love it

PR #563, opened 2018-10-26
Manually determine editor element in focus

jimlambie:
This PR manually checks the current element to determine if a toolbar icon should be active


abovedave:
Tested ✅ 

Might be worth using similar logic on the other format types e.g., quote has trouble too (also you can't toggle it, but that might be a seperate issue)

eduardoboucas:
I'm not too sure about this one, because it only partially fixes the problem and introduces new ones. Imagine the following Markdown string:

```
**using _italics_ is quite a bold move**
```

Because we're only looking at the immediate parent to determine what sort of tag we're under, it means that clicking on the word "italics" will correctly activate the italics button but not the bold.

I think the fact that the Bold button is being activated when clicking on a heading is because we're telling the editor to make headings bold: https://github.com/dadi/publish/blob/develop/frontend/components/RichEditor/RichEditor.css#L255. This doesn't have to be the case.

PR #571, opened 2018-11-02
Improve Link button in RichEditor

eduardoboucas:
(Notes in Slack)


jimlambie:
@eduardoboucas it worked pretty well for me - I wasn't sure what else to test in conjunction with the link button change.

One thing, however: if you make a word B and I, clicking "Link" removes both of those properties.

PR #572, opened 2018-11-06
Use mimetypes as filter for media reference selection

jimlambie:
This PR creates a filter to add to the URL when opening a reference field selection scree for media fields.


eduardoboucas:
Closing in favour of #573.

PR #598, opened 2018-11-14
Make RichEditor broadcast null on empty values

### LINKED ISSUE #564: Rich editor: multi-lang saves empty string
### Expected behavior

A rich editor field is left untouched, so the API returns the default field language.

### Actual behavior

It seems to save a blank `''` string, so the datasource returns empty for that field if it is not filled in the current language.

1) Enable multi-lang in Publish
2) Go to an existing document with a rich editor field
3) Switch langs (bottom left dropdown)
4) Change a non-rich editor field and save
5) Observe the API now returns that field blank in the front-end

eduardoboucas:
Closes #564.

### DIFF
--- a/frontend/components/RichEditor/RichEditor.jsx
+++ b/frontend/components/RichEditor/RichEditor.jsx
@@ -295,13 +295,19 @@ export default class RichEditor extends Component {
       text = this.getTextFromHTML(value)
     }
 
+    // An empty string is not the best representation of an empty field.
+    // We look for that case and broadcast a `null` instead.
+    let sanitisedText = text.length > 0 ?
+      text :
+      null
+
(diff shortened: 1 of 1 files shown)


eduardoboucas:
@abovedave Could you test [here](http://publish-patch-richeditor-sanitise.eb.dev.dadi.technology), please?

abovedave:
On a new article it seems fixed, I assume it won't fix an existing article because `null` fields are not updated?

On both a new and existing if I switch languages (to PT for example) and hit save with no copy in `body` it hangs and I get the error in the console:

```
code: "ERROR_VALUE_INVALID"
​​field: "body:pt"
message: "is wrong type"
```

PR #626, opened 2018-11-27
small editor improvements

abovedave:
* Bold headers
* Strike through
* Table styles: will render them nicely even though we don't support them (yet!)


abovedave:
Should we look at supporting github flavoured markdown?

eduardoboucas:
> Should we look at supporting github flavoured markdown?

I suggest we open a separate ticket for that and look into it after the public release.

PR #650, opened 2018-12-14
Improve config validation

eduardoboucas:
This PR removes most of the `server` config block from the front-end and adds some validation rules to the `apis` block, stopping the application if the legacy `credentials` block is found.


jimlambie:
Approved 🙃

PR #654, opened 2019-01-11
Various improvements for mobile

eduardoboucas:
🔍  Available for testing at http://publish-feature-mobile-improvements.eb.dev.dadi.technology/

---

# 1. Menu not visible after sign-in

Fixes an issue where, after signing in, the scroll position made it so that the nav menu wasn't visible unless the user scrolled up.

# 2. Toolbar fixed on mobile

Fixes an issue where not having enough content on the table view pushed the bottom toolbar to the middle of the screen. Now it's anchored to the bottom of the screen.

## Before

image from ios 1 1

## After

img_0683

# 3. Compress header

Reduces the size of the "Create new" button, making it on the same line as the filters and therefore reducing the amount of vertical space used.

## Before

img_0682

## After

screenshot 2019-01-11 at 14 32 21

# 4. Tighten up toolbar

For space optimisation, simplifies the pagination elements on the bottom toolbar by:

1. Removing the individual page numbers in favour of the Prev/Next buttons
1. Removing the "Showing x of y" label in favour of a new "go to page" input element, where a dropdown allows people to quickly select a page, taking advantage of the native elements from the mobile browser
1. Showing the bulk action selector only when there are documents selected, hiding it otherwise

## Before

img_0682

## After

img_0691
img_0692
img_0686

### DIFF
--- a/frontend/components/BulkActionSelector/BulkActionSelector.css
+++ b/frontend/components/BulkActionSelector/BulkActionSelector.css
@@ -1,5 +1,6 @@
 .container {
   display: flex;
+  margin-top: 10px;
   width: 100%;
 }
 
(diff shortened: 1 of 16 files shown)


jimlambie:
It all looks good, apart from having lost the pagination display

PR #659, opened 2019-01-25
Add support for custom CSS file in workspace

eduardoboucas:


app/lib/router/index.js
```diff
@@ -72,6 +72,17 @@ Router.prototype.webRoutes = function () {
       directory: path.join(process.cwd(), 'workspace')
     })(req, res, (err) => {
       if (err) {
+        // We treat this route as a special case, because index.html
+        // will always try to load it. If the user has created the
+        // file in their workspace directory, we'll serve it. If not,
+        // we don't want to return a 404 for it. We simply return a
+        // blank CSS file.
+        if (req.url === '/public/custom.css') {
```
abovedave: I'm not sure if this is needed @eduardoboucas. Don't we just need a blank `custom.css` in the public folder - that way if they're running from a direct download vs npm the solution still works?

frontend/components/Button/Button.css
```diff
@@ -22,12 +22,17 @@
...
+  transition: background-color 0.1s,
+              box-shadow 0.1s,
+              padding 0.3s,
+              background-position 0.3s;
   position: relative;
 }
 
 .button:hover {
-  box-shadow: 0 -3px 0 0 rgba(0,0,0,0.05) inset, 0 0 5px 0px rgba(0,0,0,0.15);
+  box-shadow: 10000px 0 0 rgba(0,0,0,0.1) inset,
```
abovedave: @eduardoboucas do you agree with this approach rather than the individual static shade variables?

app/lib/router/index.js
eduardoboucas: You’re right, that’s a very good point!

frontend/components/Button/Button.css
eduardoboucas: Sure, looks great.

PR #666, opened 2019-02-22
Allow for better bulk action configuration

jimlambie:
This PR moves the creation of the "Delete" option to outside the BulkActionSelector, allowing for future options to be passed in.

### DIFF
--- a/frontend/components/BulkActionSelector/BulkActionSelector.jsx
+++ b/frontend/components/BulkActionSelector/BulkActionSelector.jsx
@@ -81,27 +81,10 @@ export default class BulkActionSelector extends Component {
       modifiedActions[key] = actions[key].label
     })
 
-    let button
-
     // If the option comes with a confirmation message, display
     // ButtonWithPrompt, otherwise a standard Button is used.
-    if (selected !== ACTION_PLACEHOLDER && actions[selected].ctaMessage) {
-      button = <ButtonWithPrompt
-        accent="data"
-        disabled={(selected === ACTION_PLACEHOLDER) || actions[selected].disabled}
-        onClick={this.onApply.bind(this)}
-        promptCallToAction={actions[selected].ctaMessage}
-        promptMessage={actions[selected].confirmationMessage}
-        size="small"
-      >Apply</ButtonWithPrompt>
-    } else if (selected === ACTION_PLACEHOLDER || (actions[selected] && !actions[selected].ctaMessage)) {
-      button = <Button
-        accent="data"
-        disabled={selected === ACTION_PLACEHOLDER}
-        onClick={this.onApply.bind(this)}
-        size="small"
-      >Apply</Button>
-    }
+    const needsButtonWithPrompt = selected !== ACTION_PLACEHOLDER &&
+      actions[selected].ctaMessage
 
     return (
       <div class={containerStyle.getClasses()}>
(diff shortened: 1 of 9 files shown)


frontend/components/BulkActionSelector/BulkActionSelector.jsx
```diff
@@ -72,32 +73,49 @@ export default class BulkActionSelector extends Component {
...
+    if (selected !== ACTION_PLACEHOLDER && actions[selected].ctaMessage) {
+      button = <ButtonWithPrompt
+        accent="data"
+        disabled={(selected === ACTION_PLACEHOLDER) || actions[selected].disabled}
+        onClick={this.onApply.bind(this)}
+        promptCallToAction={actions[selected].ctaMessage}
+        promptMessage={actions[selected].confirmationMessage}
+        size="small"
+      >Apply</ButtonWithPrompt>
+    } else if (selected === ACTION_PLACEHOLDER || (actions[selected] && !actions[selected].ctaMessage)) {
```
eduardoboucas: Do we need this condition? Couldn't it be just `else`?

frontend/views/DocumentListView/DocumentListView.jsx
```diff
@@ -63,10 +63,14 @@ class DocumentListView extends Component {
...
-    const {state} = this.props  
-
-    if (actionType === BULK_ACTIONS.DELETE) {
-      this.handleDocumentDelete(state.documents.selected)
+    const {state} = this.props
+
+    switch (actionType) {
+      case BULK_ACTIONS.DELETE:
+        this.handleDocumentDelete(state.documents.selected)
+        break
```
eduardoboucas:
```suggestion
        break

```

package.json
```diff
@@ -53,6 +53,7 @@
     "connect-flash": "^0.1.1",
     "convict": "^4.0.0",
     "cookie-session": "^2.0.0-alpha.2",
+    "json-2-csv": "^3.2.0",
```
eduardoboucas: This should be part of another PR, right?

frontend/components/BulkActionSelector/BulkActionSelector.jsx
```diff
@@ -72,32 +73,49 @@ export default class BulkActionSelector extends Component {
...
+    let modifiedActions = {}
+    Object.keys(actions).forEach(key => {
+      modifiedActions[key] = actions[key].label
+    })
+
+    let button
+
+    // If the option comes with a confirmation message, display
+    // ButtonWithPrompt, otherwise a standard Button is used.
+    if (selected !== ACTION_PLACEHOLDER && actions[selected].ctaMessage) {
```
eduardoboucas: This is just a matter of personal preference, but I would create a Boolean outside of the JSX block that indicates what type of button we need and then use that as a conditional. That way the markup stays all together and the whole thing reads easier to me.

Like this:

```jsx
  render() {
    const {
      actions,
      className,
      onChange,
      selection = []
    } = this.props
    const {selected} = this.state
    const containerStyle = new Style(styles, 'container')
      .addIf('container-empty', selection.length === 0)
      .addResolved(className)
    const placeholder = 'Bulk actions'

    // Generate new actions object for the dropdown options
    let modifiedActions = {}
    Object.keys(actions).forEach(key => {
      modifiedActions[key] = actions[key].label
    })

    // If the option comes with a confirmation message, display
    // ButtonWithPrompt, otherwise a standard Button is used.
    const needsButtonWithPrompt = selected !== ACTION_PLACEHOLDER &&
      actions[selected].ctaMessage

    return (
      <div class={containerStyle.getClasses()}>
        <DropdownNative
          className={styles.dropdown}
          onChange={this.onChange.bind(this)}
          options={modifiedActions}
          placeholderLabel={placeholder}
          placeholderValue={ACTION_PLACEHOLDER}
          textSize="small"
          value={selected}
        />

        {needsButtonWithPrompt && (
          <ButtonWithPrompt
            accent="data"
            disabled={(selected === ACTION_PLACEHOLDER) || actions[selected].disabled}
            onClick={this.onApply.bind(this)}
            promptCallToAction={actions[selected].ctaMessage}
            promptMessage={actions[selected].confirmationMessage}
            size="small"
          >Apply</ButtonWithPrompt>          
        )}

        {!needsButtonWithPrompt && (
          <Button
            accent="data"
            disabled={selected === ACTION_PLACEHOLDER}
            onClick={this.onApply.bind(this)}
            size="small"
          >Apply</Button>
        )}
      </div>
    )
  }
```

package.json
jimlambie: d'oh

PR #697, opened 2019-04-18
Make DateTimePicker respect format property

### LINKED ISSUE #687: Collection field property format does not seem to affect the input field in Publish
### Expected behavior

Setting `format` as follows:

`"format": "YYYY-MM-DD",`

Should render the date picker on a DateTime field without the time option.

### Actual behavior

No change on the front end and and input seem to store the value of the field on the previous day.

Set:

`"format": "YYYY-MM-DD",`

Against a DateTime field in a collection.

eduardoboucas:
Closes #687.

### DIFF
--- a/frontend/actions/apiActions.js
+++ b/frontend/actions/apiActions.js
@@ -101,7 +101,7 @@ const applyDefaultPublishParams = fields => {
       edit: true,
       list: false
     },
-    placement: 'sidebar',
+    placement: 'main',
     section: 'General'
   }
 
--- a/frontend/components/DateTimePicker/DateTimePicker.jsx
+++ b/frontend/components/DateTimePicker/DateTimePicker.jsx
@@ -26,7 +26,12 @@ export default class DateTimePicker extends Component {
     /**
      * A callback to be fired when a new date is selected.
      */
-    onChange: proptypes.func
+    onChange: proptypes.func,
+
+    /**
+     * Whether to display controls for picking hours and minutes.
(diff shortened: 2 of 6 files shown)


davidmacp:
@eduardoboucas Selecting from the date picker or entering a date in the `DD-MM-YYYY` field the validation is triggered

Screen Shot 2019-04-25 at 11 08 59 am

jimlambie:
@eduardoboucas I've done the bare minimum for this to get it to a point where @davidmacp can continue with the testing. I've simply returned `getTime()` from the change event that is fired - you may notice that `handleChange()` is never fired at all in the FieldDateTimeEdit control. 

I've also modified the FieldDateTimeList control to apply the same format in the list view.

I suggest that once @davidmacp has a passing test in place for this, we revisit using the TextInputWithDatePicker control instead.

davidmacp:
@eduardoboucas @jimlambie Still something not quite right

Create a document with the following then `Save and go back`
Screen Shot 2019-05-08 at 8 35 57 am

Dates are displayed as so
Screen Shot 2019-05-08 at 8 36 11 am

Going back in to edit the document and the second `Date with no time` field is empty
Screen Shot 2019-05-08 at 8 36 24 am

davidmacp:
@eduardoboucas @jimlambie 

Have added tests.

However, build is now failing due to the Date & Time fields displaying as per below:

Test_Filte_1557892640 failed

PR #698, opened 2019-04-18
Fix URL of inline image select

eduardoboucas:
Fixes the URL of the inline image select view when creating new documents.

### DIFF
--- a/frontend/components/FieldString/FieldStringEdit.jsx
+++ b/frontend/components/FieldString/FieldStringEdit.jsx
@@ -161,11 +161,13 @@ export default class FieldStringEdit extends Component {
 
   handleImageSelect(position) {
     const {
+      documentId,
       name,
       onBuildBaseUrl
     } = this.props
 
-    let selectImageUrl = onBuildBaseUrl({
+    const selectImageUrl = onBuildBaseUrl({
+      createNew: !Boolean(documentId),
       referenceFieldSelect: name,
       search: {
         position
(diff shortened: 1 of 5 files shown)


davidmacp:
@eduardoboucas if I try to add an inline image through the test suite using the localhost instance that gets spun up, I am seeing the following error in console after I press `Insert image` on the select image screen

```
Uncaught TypeError: Failed to execute 'setStart' on 'Range': parameter 1 is not of type 'Node'.
    at RichEditor.value (bundle.js:5)
    at RichEditor.value (bundle.js:5)
    at FieldStringEdit.value (bundle.js:5)
    at flushMounts (bundle.js:1)
    at renderComponent (bundle.js:1)
    at Router.forceUpdate (bundle.js:1)
    at Router.routeTo (bundle.js:1)
    at bundle.js:1
    at bundle.js:1
    at dispatch (bundle.js:1)
```

jimlambie:
The problem here is probably caused by pressing Enter twice before trying to insert an image. If you press Enter once then insert an image, it works. 

The error occurs in L318 of RichEditor.jsx, where the baseNode is assigned by getting the editor's childNodes at an index that is 1 too high for the number of childNodes

jimlambie:
@eduardoboucas I added a fix in the `serialiseSelection` method, for your review when you get a moment please! 🌮

davidmacp:
Added new Inline Image test to this PR as discussed with @jimlambie

PR #701, opened 2019-04-19
Fixes to filtering

### LINKED ISSUE #699: Filtering fails when a collection schema includes a field of type Object
### Expected behavior

Filter options to be shown when a user clicks "Add filter".

### Actual behavior

The filter button triggers an error, as follows:

`Unhandled Promise Rejection: TypeError: undefined is not an object (evaluating 'fieldComponent.filterEdit')`

Screenshot 2019-04-19 at 07 04 35

This happens when a field of type Object is present in a collection schema, regardless of the settings in the `publish` block. For example:

```
    "originalObj": {
      "type": "Object",
      "publish": {
        "readonly": true,
        "display": {
          "list": false,
          "edit": false
        }
      }
    }
```

Add an object block to a collection schema.

### Package details

Using API and Publish latest builds with filestore. Running on OSX.

eduardoboucas:
This PR:

1. Fixes an issue where a collection with fields of a type that is not natively supported by Publish will cause the _Add filter_ button to throw an error

    This can be tested by navigating to the _Misc Field Testing_ collection on the preview URL, where there are 3 field types: `String`, `Color` and `Object`. Currently only `String` is filterable, so the others will simply not show on the filter selection dropdown.

1. Disables the _Add filter_ button for collections that have no filterable fields. This should be a pretty extreme case, but still worth contemplating.

    This can be tested by navigating to the _No filterable fields_ collection on the preview URL, where the only field is of type `Object`. Because that field is not filterable (not supported) at the moment, the _Add filter_ button becomes disabled.

Closes #699


davidmacp:
Re: point 1 ^

With the introduction of the `Object` and `String` fields, the `Color` field seems to have disappeared from the document edit view

Screen Shot 2019-04-25 at 8 38 28 am

jimlambie:
I've pushed a commit to this branch that correctly applies default settings:

```
display: {
  edit: true,
  list: true
},
placement: 'main',
section: 'General'
```

eduardoboucas:
Replaced by #722.

PR #716, opened 2019-05-13
Coveralls update

davidmacp:


eduardoboucas:
In #717 we removed unit tests in favour of the end-to-end suite. I'm happy to revisit once we come up with a plan to write more meaningful and useful unit tests. Until then, coverage is not a thing.

I did see [this](https://codecept.io/plugins#allure), which is something potentially useful that we could use with CodeceptJS. I'll open a separate issue for it.

Issue #688, opened 2019-04-15
Use of ellipsis on collection indexes should remove grammar immediately proceeding the ...


josephdenne:
Where an ellipsis is used on a collection index, any grammar immediately before the ... should be removed.

For example:

`Thirty years after the events of the first film,…`

<img width="1551" alt="Screenshot 2019-04-15 at 18 19 06" src="https://user-images.githubusercontent.com/150408/56152313-3c772b80-5fab-11e9-81cf-02153244ef27.png">

...should be:

`Thirty years after the events of the first film…`

PR #718, opened 2019-05-31
Fix trim with ellipsis on collection list

mstaniuk:
Fix for: #688

Steps to reproduce:
1. Go to /articles/new
2. Set new entry title to "Lorem ipsum dolor sit amet, consectetur adipisss,,,"
3. Set body to whatever
4. Save and go to /articles
5. Look at title field for new entry

Previous result:
"Lorem ipsum dolor sit amet, consectetur adipisss,…"

Result after fix:
`Lorem ipsum dolor sit amet, consectetur adipisss…`


eduardoboucas:
Thanks @mstaniuk! By the way, don't worry with the failing tests in Travis. It's expected as part of our move to React, as we didn't adjust the test suite yet.

PR #720, opened 2019-06-02
Remove trailing punctuation from truncated strings on list views

### LINKED ISSUE #688: Use of ellipsis on collection indexes should remove grammar immediately proceeding the ...

ohmoses:
Closes #688 

String values in list views are truncated with an ellipsis. This PR trims any punctuation marks which appear immediately before the ellipsis.


eduardoboucas:
Thanks @ohmoses! Please don't worry about the failing tests in Travis. It's expected as part of our move to React, as we didn't adjust the test suite yet.

PR #721, opened 2019-06-02
Route straight to edit view for single-document collections

### LINKED ISSUE #584: Single document collections
If you're building out a relatively 'static' site, it would be handy to specify that a collection only has a single document, so in Publish it skips the `list` and goes directly to the `edit` view.

A good example would be pages that have static elements made up of different fields that don't change that often e.g., homepage consisting of `headline`, `introduction` etc

ohmoses:
Closes #584 

This PR enables the flag `settings.publish.isSingleDocument` in the collection specification. This hints to Publish that the collection is supposed to only contain a single document.

With this flag set to `true`, routes `/collection` and `/group/collection` render the edit view instead of the list view, displaying either the first document in the collection, if there is one, or an empty editor for creating a new document.

Secondary save options ("Save and go back", "Save as a duplicate", "Save and create new") are removed from the save button for a single-document collection.

### Overview of code changes

App:
- Passing a render function `renderEditListSwitch()` to routes `/collection` and `/group/collection` instead of the `DocumentListView` component. The function reads the `isSingleDocument` flag and returns the proper component (`DocumentListView` or `DocumentEditView`).

DocumentEditView:
- Moved assignments to instance (`this`) properties from `render()` to a helper function `deriveInstanceProps()` so that the properties can be accessed before rendering.
- Added logic for fetching the document list for the collection.
- Refetching doc list on save/delete (L93–102) so that `this.documentId` can be properly re-read from the fetched doc list.
- Moved `documentId` to an instance property so that it can be set from the fetched doc list.

DocumentEditToolbar:
- Removed secondary actions from save button for single-doc collections.

### DIFF
--- a/app/containers/App/App.jsx
+++ b/app/containers/App/App.jsx
@@ -118,10 +118,6 @@ class Route extends React.Component {
             )
           }
 
-          if (typeof render === 'function') {
-            return render(props)
-          }          
-
           const route = {
             params: routeProps.computedMatch.params,
             path: routeProps.location.pathname,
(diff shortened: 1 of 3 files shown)


eduardoboucas:
We can close this in favour of #762, right?

PR #730, opened 2019-06-21
Allow localhost api address for backend

### LINKED ISSUE #726: Publish config to allow two locations for each API
It would be great if Publish config would allow two locations for each API.

Currently we have:

  ```"apis": [
    {
      "name": "my_cat_blog",
      "host": "http://my.cat.blog",
      "port": 80,
      ...
    }
  ]
```

But when running in Docker, we may want external traffic to go via a proxy and internal traffic to not do so, i.e. frontend -> https://my.cat.blog, backend: http://my_cat_blog_api etc.

```
"apis": [
  {
    "name": "my.cat.blog",
    "external": {
      "host": "https://api.my.cat.blog",
      "port": 443
    },
    "internal": {
      "host": "http://my_cat_blog_api",
      "port": 80
    }
    ...
  }
]
```

Pleaaaaaaase @eduardoboucas, pleaaase.

### PR DESCRIPTION
**Pull request by ohmoses**
Closes #726 

Allows to specify `localHost` and `localPort` in the Publish config—if present, the Publish backend will use this as the API address, while the frontend still uses `host` and `port`. E.g.:

```json
"api": {
    "host": "https://api.example.com",
    "port": 443,
    "localHost": "http://localhost",
    "localPort": 8081
}

### CODE REVIEW

File: server/lib/get-data-for-token.js [hunk 1]
```diff
@@ -56,63 +56,76 @@ module.exports = accessToken => {
...
+    )
+
+    if (clients.length === 0) {
+      return Promise.reject()
+    }
+
+    response.client = clients[0]
+    response.config = config.get()
+
+    const languagesPromise = request(
```
eduardoboucas: Why create two separate variables with the result of the Promise and not just:

```js
const {results: languages} = await request(
  `${apiAddress}/api/languages`,
  requestOptions
)
```

server/lib/get-data-for-token.js
```diff
@@ -56,63 +56,76 @@ module.exports = accessToken => {
...
+      .map(collection => {
+        const group = collectionGroups.get(collection.slug)
+        const href = group
+          ? `/${slugify(group)}/${collection.slug}`
+          : `/${collection.slug}`
+
+        return {
+          ...collection,
+          fields: applyDefaultPublishParams(collection.fields),
+          _publishLink: href
```
eduardoboucas: I just added two more properties to this response in #732. I think it'd be easier if we merged that one first and then you could rebase? 😇

PR #731, opened 2019-06-24
Handle API failure

### LINKED ISSUE #713: Offline API causes white screen when a session is present
If API is down and Publish is loaded you see the normal login screen and sign in fails with error `The API is not responding` however, if a session already exists and API is unresponsive, nothing appears.

I believe this should be a warning message which should also be used instead of the login page.

ohmoses:
Closes #713 

To try out the fix:
- Start Publish and API, open Publish, kill API, navigate to any collection list or document; you should see an error message.
- Restart API and click "Try again" or navigate elsewhere and back; the collection/document loads. 
- Kill API and refresh Publish; you get the error message instead of the sign in screen.

Notes/explanations:
- I tried to reuse mechanisms that exist or have existed in the code at some point, e.g. ~~`window.__api__`~~ `window.__error__` and the `api` reducer to convey the API error from the backend to the frontend.
- When you get a backend error (backend failed to connect in the first place), the "Try again" button reloads the page so that the backend gets a chance to load the data from API.
- On frontend error (API failure while navigating to a collection/document) the "Try again" button didn't trigger refetch of data—~~it navigates to the same page it's already on, so the only thing that changes is the React Router location hash key. I refetch the data if the key changes.~~ Edit: As per discussion with @eduardoboucas, the "Try again" button now flags the document or document list as dirty, forcing refetch.
- ErrorView in `App.js` has to be wrapped in Router to be able to use the Link component.
- I fixed an error where trying to delete documents while the API was down triggered a notification falsely informing of successful deletion.


app/containers/DocumentList/DocumentList.jsx
```diff
@@ -193,25 +209,17 @@ class DocumentList extends React.Component {
...
 
-      // If not, we render the SpinningWheel component.
-      return <SpinningWheel />
+    if (data.error) {
+      return typeof onError === 'function' ? onError({error: data.error}) : null
     }
 
     const {metadata, results} = data
 
-    // If there are no results to display, we call the `onEmptyList` callback
```
ohmoses: Here I took the liberty to clear out some comments which were telling me the same thing as the code, if that's okay :slightly_smiling_face:

app/containers/DocumentList/DocumentList.jsx
```diff
@@ -193,25 +209,17 @@ class DocumentList extends React.Component {
...
-    // if it is defined, or render `null` otherwise.
     if (results.length === 0) {
-      if (typeof onEmptyList === 'function') {
-        return onEmptyList({selection})
-      }
-
-      return null
+      return typeof onEmptyList === 'function' ? onEmptyList({selection}) : null
```
ohmoses: Seems quicker to read to me, what do you think? Just wanted to do a bit of minor gardening while I'm here

app/components/Button/Button.jsx
```diff
@@ -145,7 +146,7 @@ export default class Button extends React.Component {
       }
 
       return (
-        <Link className={buttonStyle.getClasses()} to={href}>
+        <Link className={buttonStyle.getClasses()} to={href} onClick={onClick}>
```
eduardoboucas: `Button` has been using `href` and `onClick` separately (i.e. never together), to decide whether to render an anchor or a button. I guess your change shouldn't have any side-effects because we check for `href` first, but I just wanted to double check with you. :)

app/components/ErrorMessage/ErrorMessage.jsx
```diff
@@ -61,7 +61,11 @@ export default class ErrorMessage extends React.Component {
       case Constants.API_CONNECTION_ERROR:
         return {
           body: (
-            <Button accent="system" href={window.location.pathname}>
+            <Button
+              accent="system"
+              href={window.location.pathname}
+              onClick={window.document.location.reload}
```
eduardoboucas: Not sure I understand this. Why are we linking to the current pathname _and_ reloading the page?

app/containers/Document/Document.jsx
```diff
@@ -43,6 +43,11 @@ class Document extends React.Component {
      */
     onRender: proptypes.func,
 
+    /**
+     * Hashed location key from React Router.
+     */
+    routeKey: proptypes.string,
```
eduardoboucas: Hmm, why does `Document` care about routing? What are the circumstances in which the content needs to be re-fetched and both a) `contentKey` hasn't changed and b) `dataIsDirty` is false?

app/containers/DocumentList/DocumentList.jsx
```diff
@@ -67,6 +73,11 @@ class DocumentList extends React.Component {
      */
     page: proptypes.number,
 
+    /**
+     * Hashed location key from React Router.
+     */
+    routeKey: proptypes.string,
```
eduardoboucas: Same question here. The reason we introduced `contentKey` was to allow the containers (i.e. `Document` and `DocumentList`) to operate independently of the view. You could have multiple on the same page. So each one gets a `contentKey`, so that they don't need to be aware of the concept of routes. Any reason to introduce this here?

server/routes/app.js
```diff
@@ -20,6 +20,7 @@ const template = ({client, config, version}) => `
       <div id="app"></div>
       <script>window.__client__ = ${JSON.stringify(client)}</script>
       <script>window.__config__ = ${JSON.stringify(config)}</script>
+      <script>window.__api__ = ${JSON.stringify(api)}</script>
```
eduardoboucas: If I understand correctly, `api` (and `window.__api__`) will contain only API errors, because the API data itself is inside `config` (and `window.__config__`). So shouldn't this be named `error` (and window.__error__) or similar? Or do you envision more data to be added to this object?

ohmoses: Done

app/containers/DocumentList/DocumentList.jsx
ohmoses: Fixed

app/containers/Document/Document.jsx
ohmoses: Fixed

app/components/ErrorMessage/ErrorMessage.jsx
ohmoses: Added an explanatory comment

app/components/Button/Button.jsx
ohmoses: Explained in a comment in `ErrorMessage.jsx`

server/lib/get-data-for-token.js
```diff
@@ -74,6 +74,7 @@ module.exports = async accessToken => {
     response.client = clients[0]
     response.config = config.get()
 
+    // run requests in parallel, then await both at once
```
eduardoboucas:
```suggestion
    // Run requests in parallel, then await both at once.
```

app/views/DocumentListView/DocumentListView.jsx
```diff
@@ -32,15 +32,16 @@ class DocumentListView extends React.Component {
...
+    const {isDeleting, error} = data
     const {isDeleting: wasDeleting} = oldData
 
     // Have we just deleted some documents?
     if (wasDeleting && !isDeleting) {
-      const message =
-        wasDeleting > 1
-          ? `${wasDeleting} documents have been deleted`
-          : 'The document has been deleted'
+      const message = error
```
eduardoboucas: I'm personally not a big fan of chaining the ternary operator, but I guess the alternative here wouldn't be much prettier? (I'm happy with it if you are.)

app/views/DocumentEditView/DocumentEditView.jsx
```diff
@@ -248,6 +248,18 @@ class DocumentEditView extends React.Component {
     )
   }
 
+  handleNetworkError() {
+    return (
+      <ErrorMessage
+        type={Constants.STATUS_FAILED}
```
eduardoboucas: Sort props alphabetically? 😇

app/views/DocumentListView/DocumentListView.jsx
```diff
@@ -158,6 +159,20 @@ class DocumentListView extends React.Component {
     )
   }
 
+  handleNetworkError() {
+    return (
+      <ErrorMessage
+        type={Constants.STATUS_FAILED}
```
eduardoboucas: Sort props alphabetically? 😇

ohmoses: done

app/views/DocumentEditView/DocumentEditView.jsx
ohmoses: done

mingard:
👏

PR #744, opened 2019-07-08
Enable string lists

### LINKED ISSUE #389: Make FieldString handle array of strings
A field which can have multiple rows of content added

```
[-] Row 1  
[-] Row 2

[+] Add new row
```

ohmoses:
Closes #389 

String fields can now have a flag `publish.list`, which makes the field display in the edit view as a **list of items** which can be added, removed, and reordered with drag-and-drop.

Behaves same as Google Keep: 
- **Enter** (**Ctrl+Enter** if `publish.multiline`) to move everything after the cursor to a new item below.
- **Delete** (if cursor is at the end of the item) and **backspace** (if cursor is at the beginning) to concatenate with next and previous item, respectively.
- **Arrow up** (if cursor at the beginning of an item) or **arrow down** (if cursor at the end) to move between items.
- Typing into the last input (the "add item" input) immediately creates the item and adds a new empty "add" input below.

### Fixed bugs
- On all plain string fields (not rich editor), the cursor would jump to the end of the input after each keystroke → fixed
- `publish.display.link` being a string (e.g. `https://twitter/{value}`) didn't work as intended; the preview link ("Open in new window") would only show if the value itself started with "http://" or "https://" → fixed

As a note, I also enabled using values that don't start with `http(s)` even without a `display.link` template; in that case, `http://` is added at the front, so that the user can write something like `github.com` and still get a preview link.

### Future to-dos
- Auto-resized multiline textarea (`heightType = 'content'`) with one line of text auto-sizes to 32px height, more than the height of the one line (20px). This is a bit ugly in lists, should be fixed eventually (but doesn't feel high priority).
- I wanted to hide the buttons (link preview, remove) unless the item is focused, dragged, or hovered over; this created a bunch of complications, but I would maybe like to revisit that at some point, after we finish the more urgent things. I left a little bit of the relevant CSS there commented out.
- If the user scrolls the page while dragging the item, it creates a bit of a funny effect, but I don't think it justifies putting the time in to fix it at the moment.
- The list logic now makes up something like half of the code of the component; possibly it could be factored out into a component of its own, just something easy like
  ```javascript
  if ((publishBlock && publishBlock.list) || Array.isArray(value)) {
    return <StringArrayEditor {...this.props} />
  }
  ```
- Tests!

list1
Screenshot_2019-07-08 Edit document DADI Publish
Screenshot_2019-07-08 Edit document DADI Publish(1)


ohmoses:
Oh, while I was touching the code, I changed a couple of instances of `someFn.call(this, ...)` to `someFn(...)`, it evaluates to the same thing when the function is not called as a method. Hope that's okay.

app/components/FieldString/FieldStringEdit.jsx
```diff
@@ -113,11 +125,77 @@ export default class FieldStringEdit extends React.Component {
...
     super(props)
 
+    this.cursor = null
+    this.dragBottomLimit = null
+    this.dragStart = {}
+    this.inputRefs = []
+    this.itemRefs = []
+    this.listContainerRef = null
+
+    this.mousePositionListener = ({clientY: currentY}) => {
```
eduardoboucas: I understand that you're declaring `this.mousePositionListener` and `this.dragEndListener` as instance variables in the constructor so that you have a single reference to pass to `addEventListener`/`removeEventListener`, but I kinda feel it makes the constructor a bit messy. Could you perhaps declare them as class methods and then declare an instance variable that simply binds them?

```js
constructor(props) {
  this.mousePositionListener = this.handleMousePositionChange.bind(this)
}
```

app/components/FieldString/FieldString.css
```diff
@@ -28,15 +28,76 @@
...
+.dragged .icon-drag {
+  opacity: 1;
+}
+
+.link-preview {
+  white-space: nowrap;
+  margin-right: 5px;
+}
+
+/* .remove-button,
```
eduardoboucas: Can we remove this, or will it be used later on?

app/components/FieldString/FieldStringEdit.jsx
```diff
@@ -133,31 +211,217 @@ export default class FieldStringEdit extends React.Component {
...
     }
   }
 
+  handleListInputChange(index, e) {
+    const {onChange, value: oldValueArray} = this.props
+
+    if (typeof onChange !== 'function') return
+
+    const newValueArray = Array.isArray(oldValueArray)
+      ? oldValueArray.slice()
```
eduardoboucas: Are you calling `slice` with no arguments to get a new array? If so, can we use `Array.from` instead?

app/components/TextInput/TextInput.jsx
```diff
@@ -147,15 +152,21 @@ export default class TextInput extends React.Component {
 
   componentDidMount() {
     this.adjustHeightIfNeeded()
+
+    const {inputRef} = this.props
+
+    typeof inputRef === 'function' && inputRef(this.base)
```
eduardoboucas: I'm personally not a fan of this type of conditional, as it's not immediately clear the code is branching off. Plus, we use `if (typeof x === 'function')` in numerous places across the application to check whether a prop or callback are defined, so I'd rather keep with that format whenever possible.

eduardoboucas:
Added some minor style comments. Haven't tried the functionality yet. I think I'll try to change the CI box to use the repo API first. 🤞

app/components/TextInput/TextInput.jsx
ohmoses: Fixed

app/components/FieldString/FieldStringEdit.jsx
ohmoses: Fixed

app/components/FieldString/FieldString.css
ohmoses: I did want to use it later, but realistically, it will not be so soon, so I took it out

app/components/FieldString/FieldStringEdit.jsx
ohmoses: Done!

ohmoses:
@davidmacp Awesome, thanks! :tada:

PR #757, opened 2019-07-21
Fix a few minor datetime bugs

ohmoses:
I put a few fixes together into a single PR, because they're tiny.

### Bugs fixed:

1. **When manually typing a date into a date input field, the app would go into an infinite loop and freeze.**
1. When manually typing the date, after going from an invalid value to a valid one, the cursor would jump to the end of the input.
1. The date input field would not allow deleting the value completely.

Also, I fixed an unrelated potential bug in `TextInput` (in fortunately otherwise unused code):
```javascript
const {oldHeightType} = oldProps
```
changed to
```javascript
const {heightType: oldHeightType} = oldProps
```

### DIFF
--- a/app/components/TextInputWithDatePicker/TextInputWithDatePicker.jsx
+++ b/app/components/TextInputWithDatePicker/TextInputWithDatePicker.jsx
@@ -141,26 +141,24 @@ export default class TextInputWithDatePicker extends React.Component {
     const {value} = event.target
     const newDate = new DateTime(event.target.value, format)
 
-    if (value.length > 0) {
-      if (value.length === format.length && newDate.isValid()) {
-        if (typeof onChange === 'function') {
-          const utcDate = newDate.getDate({toUTC: true})
+    if (value.length === format.length && newDate.isValid()) {
+      if (typeof onChange === 'function') {
+        const utcDate = newDate.getDate({toUTC: true})
 
-          onChange(utcDate.getTime())
-        }
-
-        this.setState({
-          internalValue: null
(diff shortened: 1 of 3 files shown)


app/components/TextInputWithDatePicker/TextInputWithDatePicker.jsx
```diff
@@ -141,26 +141,24 @@ export default class TextInputWithDatePicker extends React.Component {
     const {value} = event.target
     const newDate = new DateTime(event.target.value, format)
 
-    if (value.length > 0) {
```
ohmoses: Make sure you have 'hide whitespace changes' checked, otherwise the diff looks more complicated than it is :slightly_smiling_face:

app/lib/datetime.js
```diff
@@ -13,52 +13,32 @@ export default class DateTime {
...
+    const intDate = parseInt(date, 10)
+
+    // Unix timestamp in a string.
+    if (intDate.toString() === date) return new Date(intDate)
+
+    try {
+      const parsedDate =
+        // Fecha requires a format. If no format is specified, try ISO.
+        fecha.parse(date, format || FORMAT_ISO8601) ||
+        // If format is specified but parsing fails, the value has probably been
```
eduardoboucas:
```suggestion

        // If format is specified but parsing fails, the value has probably been
```

ohmoses: Nope, sorry, prettier doesn't like that :grin:

PR #758, opened 2019-07-23
Fix sorting docs in ReferenceSelectView

### LINKED ISSUE #642: Sorting in reference select view
### Expected behavior

When selecting a document for a reference field, clicking on column headings to sort should sort the documents available for selection

### Actual behavior

When selecting a document for a reference field, clicking on column headings to sort goes back to the parent collection and applies the sort to that

### To reproduce

Use a collection with a reference field
Create a new document
Click "Select new whatever"
Click a column heading to sort

ohmoses:
Closes #642

### DIFF
--- a/app/components/DocumentTableList/DocumentTableList.jsx
+++ b/app/components/DocumentTableList/DocumentTableList.jsx
@@ -120,8 +120,8 @@ export default class DocumentTableList extends React.Component {
   }
 
   handleTableSort(value, sortBy, sortOrder) {
-    const {onBuildBaseUrl} = this.props
-    const link = onBuildBaseUrl({
+    const {buildSortUrl = this.props.onBuildBaseUrl} = this.props
+    const link = buildSortUrl({
       search: {
         order: sortOrder,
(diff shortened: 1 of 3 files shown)


app/components/DocumentTableList/DocumentTableList.jsx
```diff
@@ -120,8 +120,8 @@ export default class DocumentTableList extends React.Component {
   }
 
   handleTableSort(value, sortBy, sortOrder) {
-    const {onBuildBaseUrl} = this.props
-    const link = onBuildBaseUrl({
+    const {buildSortUrl = this.props.onBuildBaseUrl} = this.props
```
eduardoboucas: Can you clarify this part? Is this so that sorting will use `buildSortUrl` and fall back to `onBuildBaseUrl`?

ohmoses: Yes, exactly

eduardoboucas:
Is this good to go?

PR #762, opened 2019-08-04
Add support for single-document collections

### LINKED ISSUE #584: Single document collections

ohmoses:
Closes #584 

If `settings.publish.isSingleDocument` is set to `true` in the collection schema, Publish will show the document editor (instead of the document list) on route `/collectionName` (or `/groupName/collectionName`). If there are multiple documents in the collection, Publish will operate on the first one.

I made a small change to enable `connectRedux` to accept a `mapState` function, so that in `ReferenceSelectView`, `DocumentListView`, and `DocumentEditView`, I could move the logic for creating `contentKey`s, `documentId`, etc., to `mapState` and pass them to the components.


app/views/DocumentEditView/DocumentEditView.jsx
```diff
@@ -28,28 +28,27 @@ class DocumentEditView extends React.Component {
...
-    const {documentId} = route.params
-    const document = state.document[this.contentKey] || {}
-    const oldDocument = oldState.document[this.contentKey] || {}
+    const document = state.document[contentKey] || {}
+    const oldDocument = oldState.document[contentKey] || {}
     const isSaving = (oldDocument.saveAttempts || 0) < document.saveAttempts
 
     // Are there unsaved changes?
     if (
+      !document.isLoading &&
```
ohmoses: This fixes a minor bug: if there previously were unsaved changes, but now there are not, then in the short amount of time before the document is loaded (with `wasLoadedFromLocalStorage` being false), the previous version of the document (with `wasLoadedFromLocalStorage` true) is grabbed from Redux, which causes the "unsaved changes" notification to fire.

ohmoses:
A point for discussion: I removed the "Delete" button, because it didn't make much sense here, but maybe there should be something like "clear all fields" button?

app/views/ReferenceSelectView/ReferenceSelectView.jsx
```diff
@@ -195,25 +169,34 @@ class ReferenceSelectView extends React.Component {
...
     const pageNumber = parsedPage.toString() === page ? parsedPage : undefined
 
-    // Getting the schema for the parent collection.
-    const collection =
-      collectionName === Constants.MEDIA_COLLECTION_SCHEMA.slug
-        ? Constants.MEDIA_COLLECTION_SCHEMA
-        : api.collections.find(collection => {
-            return collection.slug === route.params.collection
-          })
+    if (!(isNewDocument || isSingleDocument || documentId)) {
```
eduardoboucas:
```suggestion
    if (!isNewDocument && !isSingleDocument && !documentId)) {
```

... is easier on the brain, I think?

app/views/ReferenceSelectView/ReferenceSelectView.jsx
```diff
@@ -384,6 +366,21 @@ class ReferenceSelectView extends React.Component {
         }}
       />
     )
+
+    if (isSingleDocument) {
+      // In this case the documentId is not in the url, so we have to fetch the
```
eduardoboucas:
```suggestion
      // In this case the documentId is not in the URL, so we have to fetch the
```

🤓

app/views/ReferenceSelectView/ReferenceSelectView.jsx
```diff
@@ -449,6 +446,54 @@ class ReferenceSelectView extends React.Component {
...
+  } = ownProps
+
+  const collection =
+    params.collection === Constants.MEDIA_COLLECTION_SCHEMA.slug
+      ? Constants.MEDIA_COLLECTION_SCHEMA
+      : state.app.config.api.collections.find(collection => {
+          return collection.slug === params.collection
+        })
+
+  const isNewDocument = /\/new\/?/.test(path) && !params.documentId
```
eduardoboucas: Is this a good check? It correctly matches `/articles/new`, but also matches `/newyork/12345`.

app/views/DocumentListView/DocumentListView.jsx
```diff
@@ -229,8 +214,36 @@ class DocumentListView extends React.Component {
...
+              <Main>
+                <SpinningWheel />
+              </Main>
+            </Page>
+          )}
+          onNetworkError={this.handleNetworkError.bind(this)}
+          onRender={({documents}) => (
+            <DocumentEditView
+              {...this.props}
+              isSingleDocument
```
eduardoboucas: Sort props alphabetically, please. 🤓

app/containers/App/App.jsx
```diff
@@ -280,95 +331,17 @@ class App extends React.Component {
...
-          <Route
-            isSignedIn={isSignedIn}
-            exact
-            component={ReferenceSelectView}
-            config={state.app.config}
-            mustBeSignedIn
-            path={`/:group${REGEX_SLUG}/:collection${REGEX_SLUG}/:documentId${REGEX_DOCUMENT_ID}/select/:referenceField/:page${REGEX_NUMBER}?`}
-          />
+          {collectionRoutes.map(({path, component}) => (
+            <Route
```
eduardoboucas: Do you mind sorting the props alphabetically here, please? (I know they weren't before.)

app/views/ReferenceSelectView/ReferenceSelectView.jsx
```diff
@@ -449,6 +446,54 @@ class ReferenceSelectView extends React.Component {
...
+  } = ownProps
+
+  const collection =
+    params.collection === Constants.MEDIA_COLLECTION_SCHEMA.slug
+      ? Constants.MEDIA_COLLECTION_SCHEMA
+      : state.app.config.api.collections.find(collection => {
+          return collection.slug === params.collection
+        })
+
+  const isNewDocument = /\/new(\/|$)/.test(path) && !params.documentId
```
eduardoboucas: @ohmoses is there ever a case where checking for `params.documentId` is not enough to determine whether we're creating a new document?

PR #781, opened 2019-09-23
Update fonts and paddings

ohmoses:


eduardoboucas:
So much better!

This field suddenly looks weird, though. Maybe its height needs to be adjusted as per the new font size?

Screenshot 2019-09-23 at 19 13 30

PR #783, opened 2019-09-25
Add loading animation

eduardoboucas:


eduardoboucas:
I've reinstated the overlay and made it so that the navigation isn't covered by the overlay.

@ohmoses could you check?

PR #787, opened 2019-09-27
Make rich editor use ReferenceSelectView for inserting media

eduardoboucas:


eduardoboucas:
Good catch! That was indeed a bug. Should be fixed now.

PR #788, opened 2019-09-29
Reskin mobile document section navigation

ohmoses:
I go against the designs in that the dropdown is not fullscreen; I vaguely remember from a previous job that trying to prevent the page from scrolling when a dropdown/modal is showing creates problems in certain mobile browsers and it's a mare to make it work. It's up for discussion, though.

Changed the mobile filter dropdown in the same way.


eduardoboucas:
Yep, that’s sensible. I would change it so that the tabs (and probably even the filters) actually push the content down, rather than having them float over it.

PR #808, opened 2019-10-11
Remove media card hover effect if checkbox hovered

ohmoses:
+ fix that `props.onHover is not a function` error in document table

### DIFF
--- a/app/components/Table/TableRowCell.jsx
+++ b/app/components/Table/TableRowCell.jsx
@@ -34,10 +34,8 @@ export default class TableRowCell extends React.Component {
     return (
       <td className={cellStyle.getClasses()}>
         {React.Children.map(this.renderChildren(), child => {
-          // Only pass onHover to field components.
-          return child.type &&
-            child.type.name &&
-            child.type.name.startsWith('Field')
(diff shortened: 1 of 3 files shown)


app/components/Table/TableRowCell.jsx
```diff
@@ -34,10 +34,8 @@ export default class TableRowCell extends React.Component {
     return (
       <td className={cellStyle.getClasses()}>
         {React.Children.map(this.renderChildren(), child => {
-          // Only pass onHover to field components.
-          return child.type &&
-            child.type.name &&
-            child.type.name.startsWith('Field')
+          // Only pass onHover to React components.
+          return typeof child.type === 'function'
```
eduardoboucas: I'm pretty sure we were getting errors because `child` didn't have a `type` property, so I'd leave the first check in there.

ohmoses: No problem, if `child` doesn't have a `type` property, the typeof will be `"undefined"`

eduardoboucas: Of course! 🤦‍♂

PR #809, opened 2019-10-13
Change layout of boolean field

ohmoses:


eduardoboucas:
I think the label needs to be flush with the toggle. The toggle needs to move left a bit:

Screenshot 2019-10-13 at 21 09 14

ohmoses:
> I think the label needs to be flush with the toggle. The toggle needs to move left a bit:
> 
> Screenshot 2019-10-13 at 21 09 14

@eduardoboucas Done

PR #822, opened 2019-11-11
Refactor list view

ohmoses:


app/components/MediaList/MediaList.jsx
```diff
@@ -0,0 +1,144 @@
...
+
+    this.state = {
+      mediaListMode: 'grid'
+    }
+  }
+
+  render() {
+    const {
+      documents,
+      isFilteringSelection,
```
eduardoboucas: Sort alphabetically, please? 😇

app/components/MediaList/MediaList.jsx
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * The name of the field currently being used to sort the documents.
+     */
+    sort: proptypes.string
+  }
+
+  constructor(props) {
+    super(props)
+
+    this.renderFieldMediaItem = ({document}) => (
```
eduardoboucas: I'm not sure I like the idea of having this in the constructor. I know that it's a single JSX node, but I think it creates a weird precedent. Would you mind moving this to a function?

app/components/MediaList/MediaList.jsx
```diff
@@ -0,0 +1,144 @@
...
+      <FieldMediaItem config={this.props.config} isList value={document} />
+    )
+    this.updateMediaListMode = mediaListMode => this.setState({mediaListMode})
+
+    this.state = {
+      mediaListMode: 'grid'
```
eduardoboucas: I think I would rename this to just `listMode`, since we're now inside a component that is specific to media.

